PR #7, opened 2018-12-23
[CAMEL-12998] Add responsive CSS

fpapon:
Update design with responsive CSS based on Bootstrap 4.


zregvart:
Thanks for the pull request. I really appreciate your ongoing efforts in this. I would prefer if we used just the grid system or just custom breakpoints flebox css from one of the much smaller frameworks. This increases base.css from 815 bytes to 75Kb.

Would you consider basing this work on [min](http://mincss.com/) for example?

fpapon:
I don't get the full Bootstrap classes but picked just some I think it could be very helpful to design web pages like padding, margin, col, card, text transformation... I was thinking that it could help other user in the contribution of the website to make pages faster and easier :)

I take a look on MinCSS and it missing some useful classes to have a nice style and I have to add some of them manually so I need to increase the size of the css too.

Btw, if you consider that it's necessary to user MinCSS, I will do it ;)

zregvart:
I'm back, happy holidays!

I'd personally stay away from Bootstrap due to size of unused CSS I think it'll end up costing us. Despite my fixation with CSS size let's have this merged and work on getting the CSS smaller over time.

I propose the following, to maintain the ability to rebase easily on the default Antora UI let's move our additions to a separate `.css` file (say `responsive.css` in this case) and `@import` it `site.css`. This way we can have this change merged and refactor/optimize it over time so it only has the minimum CSS we're using.

Over time I'd like to maintain as small subset of changes over the default Antora UI as possible so it's easier to bring in improvements made in there.

If our pages not built by Antora end up having a bunch of CSS we can consider splitting the CSS for those pages out of the Camel Antora theme and add it to the website directly.

zregvart:
Awesome, thanks @fpapon!

PR #17, opened 2019-02-18
[CAMEL-12998] Add articles and books pages

fpapon:


zregvart:
I think we're missing the Books menu item, but that can be added in subsequent PR.

PR #22, opened 2019-03-08
[CAMEL-11501] Update README

christophd:
This PR updates both root README and antora-ui-camel README in order to describe the build process and where to add someones changes.

I had to remove pnp install config as it always failed on my machine. I am sure this is because of some newer node and yarn versions on my machine. Plain old yarn with `node_modules` works fine though.

Also introduced an optional Maven build process because of these reasons:
- Build is more stable as it always uses the same versions of yarn and node
- Build combines Antora UI theme build and Hugo site generation in one single build
- Users not familiar with node, yarn, gulp and so on can easily build the UI (only Maven required)

The README describes both ways to build the website with and without Maven.


zregvart:
We control the version of yarn via `.yarnrc`, so no matter what version of yarn you have installed the version we have in the repository is used. See:

https://github.com/apache/camel-website/blob/d98c2301320b0ebe857a0d304d33b23cbfa4966a/.yarnrc#L6

https://github.com/apache/camel-website/blob/d98c2301320b0ebe857a0d304d33b23cbfa4966a/antora-ui-camel/.yarnrc#L6

I would like not to have another build tool in the mix. Having Node and Yarn should suffice. Yarn doesn't currently support Node 11, and that's why we encounter issues using it. Note that on the Node site version 10 is LTS and "Recommended For Most Users", so I think it's a good decision to stick with that version. Adding Maven to the mix is just another tool that we have to take on and maintain, and I'd like to keep the project as simple as it can be. Using PnP is much more beneficial than using the latest Node version. With PnP we can cache Node modules outside of the git repository which drastically speeds up the build on CI where the build runs on different nodes without any opportunity for caching, so all Node modules are re-downloaded/build.

To use a fixed Node version I believe [nvm](https://github.com/creationix/nvm) is often used. Antora theme already contains `.nvmrc` to do so:

https://github.com/apache/camel-website/blob/460e547b3c4dba09e3452285f06ffbcad1bc0046/antora-ui-camel/.nvmrc#L1

I'd rather see an investment in existing build toolset Node/Yarn/gulp than adding another build tool.

christophd:
I can confirm that the PnP errors were related to me using Node v11 instead of Node v10. With Node v10 everything works out fine. I will add this to the README as well.

So I will bring back PnP and I will also remove the Maven build from this PR and perhaps open a new one with that Maven build option (as I still see some benefit from it that we can discuss in another PR).

Let me fix these things and force push to this PR

christophd:
@zregvart please have another look. Force pushed my changes according to your feedback to this PR

PR #23, opened 2019-03-12
chore: Add Maven build option

christophd:
This PR adds a Maven build option to build the website content

Using Maven in my opinion provides some benefits:
- Combines Antora UI and Hugo website build, where Antora UI theme is automatically built first
- Easy to use for users not familiar with Node and yarn
- No global installation of Node tooling required
- Users having incompatible Node version on their machine do not need to manage multiple Node versions on the host
- `mvn clean` removes all build output folders for a clean environment build

The Maven build option is additional to the normal build using 'Node' and 'yarn' directly. Both build options can coexist and produce the exact same output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,14 @@ Tools used to generate the website:
    Asciidoc documents from different sources in the [Camel](https://github.com/apache/camel) 
    and [Camel K](https://github.com/apache/camel-k) repositories where user manual and component 
    reference documentation resides and renders them for inclusion in this website.
+ - [Maven](https://maven.apache.org/) (optional) a build tool used to run the complete website generating process
 
-## Building the website
+## Build with Node and yarn
 
-You can build the website locally using the tools `Node.js` and `yarn`.
+You can build the website locally using the tools `Node.js` and `yarn`. 
+
+If you can not use these tools on your local machine for some reason you can also build the website using Maven as 
+described in section ["Build with Maven"](#build-with-maven).
 
 ### Preparing the tools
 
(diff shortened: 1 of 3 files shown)


site.yml
```diff
@@ -15,6 +15,9 @@ content:
       branches: antora
       start_path: docs
 
+runtime:
+  cache_dir: ./.cache/antora
```
zregvart: This migh prevent us from using build wide cache on CI.

README.md
```diff
@@ -19,10 +19,74 @@ Tools used to generate the website:
    Asciidoc documents from different sources in the [Camel](https://github.com/apache/camel) 
    and [Camel K](https://github.com/apache/camel-k) repositories where user manual and component 
    reference documentation resides and renders them for inclusion in this website.
+ - [Maven](https://maven.apache.org/) (optional) a build tool used to run the complete website generating process
 
-## Building the website
+## Build with Maven
```
zregvart: If building with Maven is optional it should be the last mentioned method here. First method should be the using Node and Yarn.

site.yml
christophd: Yes, true. We do not need it I just thought it might be good to have everything within project scope

README.md
christophd: Agreed! Let me change the ordering

site.yml
christophd: How about parallel builds on CI do they share the same cache? Might this get us into trouble?

zregvart: I think not with yarn and PnP, it uses hashes in module cache and the cache is idempotent.

christophd: The Antora cache is more about the fetched github repos for Camel and Camel K but these will also not get us into trouble when used from multiple builds I guess

zregvart: Oh right, forgot we were talking about Antora cache, we place that cache within the build workspace, which is used by only one build at the time.

https://github.com/apache/camel-website/blob/3b8f70a694952a960505398e16affcd43a17fe83/Jenkinsfile#L39

Similar for GitHub Actions WIP

christophd: I see so no need for the local cache

christophd:
@zregvart force pushed to this PR: removed local cache and moved Maven build option to the end of the README

PR #25, opened 2019-03-13
fix: Use same top level navigation in Hugo and Antora generated pages

christophd:
Antora generated pages were not using the same top level navigation as the Hugo generated site. I manually added the navigation items to the Antora sources as a quickfix

In general I think it would be a good idea to somehow generate the Antora header navigation from Hugo `config.toml` information in a preprocessing step.

### DIFF
--- a/antora-ui-camel/src/partials/header-content.hbs
+++ b/antora-ui-camel/src/partials/header-content.hbs
@@ -1,7 +1,7 @@
 <header class="header">
-  <nav class="navbar">
+  <nav class="navbar shadow border-bottom">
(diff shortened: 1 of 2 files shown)


christophd:
Created a new task in Jira regarding auto synchronization of navigations https://issues.apache.org/jira/browse/CAMEL-13323

zregvart:
Awesome, I've checked out the [preview environment](https://builds.apache.org/job/Camel.website/job/PR-25/Preview/) looks good.

Thanks :100:

PR #26, opened 2019-03-13
Updated packages when building the project

Nayananga:
When I was building the project some errors fixed by these new packages.


zregvart:
@Nayananga thanks for this.

Can you provide more detail on what failed. Thus far we had no issues that required I can't find that any dependency we currently have requires `gulp-autoprefixer` or `gulp-browserify` and you haven't added any code that necessitates those.

We like to use [tilde ranges](https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1) (`~`) which allows us to get minor version upgrades (bugfixes) rather than [caret ranges](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) which can lead to upgrades of major versions which we might not be what we want.

Nayananga:
I think I may have miss guided the Readme, I completely removed and rebuilt the project with no issues. And I noticed ReadME has change comparing to my fork. :)

zregvart:
OK, cool, glad you got it running, and thanks for raising this pull request. I'll close it as it doesn't seem to be needed.

Thanks!

PR #34, opened 2019-05-02
CAMEL-13470: Publish security advisories

zregvart:
Adds two layouts for `security` and `security-advisory` and a archetype
for `security-advisory`. Security advisories are now served from
`/security` (instead of `/security-advisories.data`) so a `.htaccess`
file was added to redirect from previous URL to the current one.

README was updated to show how to use the Hugo archetype to publish a
new security advisory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@ node
 target
 node_modules
 public
-static
+static/*
+!**/.htaccess
 
 .idea
 *.iml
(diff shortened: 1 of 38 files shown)


zregvart:
@oscerd can you take a look at this from workflow/organization point of view.

There is a preview on the branch job at https://builds.apache.org/view/C/view/Apache%20Camel/job/Camel.website/job/issue%252FCAMEL-13470/Preview/

PR #44, opened 2019-05-31
 Sync `yarn.lock` package versions, into package.json

Nayananga:
Since we are not using npm package.json had outdated in both camel and antora, so I synced yarn.lock into package.json
I think it will be better if readme update to install package syncyarnlock globally and sync it when we do a yarn install or yarn upgrade.


Nayananga:
Thank You Mr.Zoran,
Closing pull request.

PR #49, opened 2019-07-05
fixes for page not found errors

Nayananga:
Sorry for not being active in this week, My new semester began since I could`t do much. Hopefully do better in this weekend


zregvart:
Thank you :)

PR #73, opened 2019-08-14
Redesign

zregvart:
With this there's a new front page design (still a bit work in progress), I started by consolidating CSS and ended up redesigning...

This consolidates the CSS used between Antora and the Hugo built parts very much a work in progress. The end result would be same exact styling that we have in the Antora bits in the Hugo built bits and also integration of only the used bits in Bootstrap into the CSS we have.

For this markup also needed to change and some visual differences will be noticed between current and the proposed changes here.

A lot of the repeating layouts were removed so that the default layout is used instead, thus reducing some of the repetition and easing future maintenance.


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/view/change-requests/job/PR-73/Preview/index.html).

This is how the front page looks with this change:

Screenshot_2019-08-14 Home page - Apache Camel(1)

Feedback welcome.

davsclaus:
Great work Zoran

A few comments

1)
I would remove the section with:

INTEGRATED WITH APACHE CAMEL
The following projects can leverage Apache Camel as a routing and mediation engine

... as this is very old content

2)
At the section where Camel and Camel Quarkus is mention, then I would add Apache Camel before them, so we have all 3, otherwise its actually a bit unclear.

3)
Blueprint -> OSGi Blueprint
And also add Standalone

And we should likely update it a bit as it mentions OSGi first.

And I wonder if Portability should be named Runtimes ?

4)
I would put the new Camel logo more prominently in the top

5)
Remove the last dot from the headline, Integration you want.

PR #74, opened 2019-08-15
Add search for Hugo (more polishing need to do) 

Nayananga:
I need to set CSS for search and each search result should able to link to the corresponding page when clicked. I would like to know is this approach acceptable?


zregvart:
Thanks for this. Can you move the bits that refactor the microdata/seo into a separate PR, from what I understand it's not a requirement for the search to function.

The JavaScript bits need to go to the Antora UI theme, the reason for that is that there's a gulp pipeline that minifies and merges all JavaScript into a single JavaScript file that also has cache bursting filename. This is important because otherwise changes to the script code might not be reflected in the browser as it's using cached version.

.env
```diff
@@ -0,0 +1,5 @@
+ALGOLIA_APP_ID={{ YOUR_APP_ID }}
+SEARCH_ONLY_API_KEY={{ SEARCH_ONLY_API_KEY }}
+ALGOLIA_ADMIN_KEY={{ YOUR_ADMIN_KEY }}
+ALGOLIA_INDEX_NAME={{ YOUR_INDEX_NAME }}
```
zregvart: I assume we'd need to add this to the build, not sure if the `$YOUR_...` bits are needed, we can simply ensure that the correct secrets are loaded to the environmet under `$ALGOLIA...`. Can you comment on which of these are secret and which are public? I'm thinking that the `ALGOLIA_INDEX_NAME` might be public.

layouts/_default/list.algolia.json
```diff
@@ -0,0 +1,54 @@
+{{/* Generates a valid Algolia search index */}}
+{{- $hits := slice -}}
+{{- $section := $.Site.GetPage "section" .Section }}
+{{- $validVars := $.Param "algolia.vars" | default slice -}}
+{{- $validParams := $.Param "algolia.params" | default slice -}}
+{{/* Include What Pages? */}}
+{{ range $i, $hit := .Site.AllPages }}
```
zregvart: I'm assuming that this only goes through the Hugo generated pages, not the Antora generated ones, am I correct?

layouts/partials/header.html
```diff
@@ -27,6 +27,10 @@
     <meta name="keywords" content="{{ .Keywords }}">
     {{ end }}
     <link rel="stylesheet" href="{{ path.Join "_" (index .Site.Data "rev-manifest" "css/site.css") | relURL }}">
+    <!-- search-bar content -->
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.1.0/themes/algolia.css" />
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/selectize@0.12.6/dist/css/selectize.css">
+    <link rel="stylesheet" href="/css/search.css">
```
zregvart: These stylesheets need to go to the Antora UI theme.

static/css/search.css
```diff
@@ -0,0 +1,28 @@
+h1 {
```
zregvart: This file needs to go to the Antora UI theme.

layouts/partials/footer.html
```diff
@@ -42,19 +42,11 @@
...
-          ],
-          "logo": "{{ .Site.BaseURL }}_/img/logo-d.svg",
-          "description": "Apache Camel ™ is a versatile open-source integration framework based on known Enterprise Integration Patterns. Camel empowers you to define routing and mediation rules in a variety of domain-specific languages, including a Java-based Fluent API, Spring or Blueprint XML Configuration files, and a Scala DSL."
-        }
-    </script>
-    {{ partial "breadcrumbs.html" . }}
+    <script src="https://cdn.jsdelivr.net/npm/jquery@3.3.1/dist/jquery.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/selectize@0.12.6/dist/js/standalone/selectize.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/algoliasearch@3.30.0/dist/algoliasearchLite.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/instantsearch.js@beta"></script>
```
zregvart: I'd prefer if we vendor all dependencies. Not sure how much more do these dependencies add to the page weight. In general I would try to minimize the 3rd party dependencies in the website runtime. JQuery is 85KB and in total we're currently serving 28KB, so that would be almost 3x increase in JavaScript. I know both numbers aren't that big, but currently we have sub second time to first meaningful paint and I'm afraid of putting additional weight on the page might jeopardize that.

zregvart:
The build is currently failing with:
```
[2019-08-15T16:38:45.078Z] $ antora --require ./menu.js --clean --fetch site.yml
[2019-08-15T16:40:36.833Z] error: The partial seo/main could not be found
```

layouts/_default/list.algolia.json
Nayananga: yes, it is for now :)

.env
Nayananga: Hello Mr.Zoran I think APP_ID and INDEX_NAME can be public ill put exacts values into them, I`m not sure whether SEARCH_ONLY_API_KEY public or not but ADMIN_KEY is definitely not public.

layouts/_default/list.algolia.json
Nayananga: I've added antora indices using docker as per in this guide https://gitlab.com/antora/antora-ui-default/issues/44#note_68053321

zregvart:
I think I'll pick bits from this into separate pull requests. I've read a bit about [Algolia docsearch](https://community.algolia.com/docsearch/).

I'm not sure if we should [run our own crawler job](https://community.algolia.com/docsearch/run-your-own.html) or let Algolia index our documentation on it's own.

Either way I think we need to include [just the JavaScript bits](https://community.algolia.com/docsearch/dropdown.html) for the docusearch dropdown.

So, perhaps as a first start we just add the docsearch dropdown and let Algolia index the site. I'll pick the changes from here needed to do that.

Nayananga:
Hello Mr. Zoran then ill try to customize the appearance of the search bar. Thank you

PR #75, opened 2019-08-20
chore: use terser instead of uglify

zregvart:
@Nayananga I've noticed in https://github.com/apache/camel-website/pull/74/commits/0ca2aad865ac7633c65439db5a0cba66c03b24a2 you've replaced uglify with terser. I think that change can be merged separately. I also think that we should split the non-search related work from the #74. That's why I created this PR based on your work there. Can you have a look and see if it includes everything needed to move from uglify to terser?

### DIFF
--- a/antora-ui-camel/gulp.d/tasks/build.js
+++ b/antora-ui-camel/gulp.d/tasks/build.js
@@ -20,7 +20,7 @@ const postcssVar = require('postcss-custom-properties')
 const rename = require('gulp-rename')
 const rev = require('gulp-rev')
 const template = require('gulp-template')
-const uglify = require('gulp-uglify')
+const terser = require('gulp-terser')
 const vfs = require('vinyl-fs')
 
 module.exports = (src, dest, preview) => () => {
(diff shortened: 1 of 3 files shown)


Nayananga:
Hello Mr. Zoran I had to user terser because uglify gave me error when parsing search.js. The reason it was unable to parse functions (as i remember)
And thats all for me to move from uglify to terser. 
Thank you. I didnt knew how to make 2 separate PRs at the same time. Sorry for that.

Nayananga:
Hello Mr. Zoran can you include [https://github.com/apache/camel-website/pull/74/commits/1a147dc99611d3adaaf83390c2108609ab1c07be](url) this commit as well?

zregvart:
> Hello Mr. Zoran can you include this commit as well?
> [1a147dc](https://github.com/apache/camel-website/commit/1a147dc99611d3adaaf83390c2108609ab1c07be)

So let's do that piecemeal, I'll merge this to change uglify to terser, then I can pick those changes in a separate pull request. It's easier to work in smaller batches and smaller PRs.

And you can create as many PRs as you need, the issue is that sometimes PRs can depend on each other. I wouldn't worry too much about it though...

PR #79, opened 2019-08-20
config.toml should include params inorder to perform microdata scripts correctly

Nayananga:
I`ve mistakenly forgotten to include this change into "microdata" section from "add search " section. But this change is necessary in order to perform microdata scripts correctly. Sorry for the mistake


zregvart:
No worries, thanks for the fix :)

PR #83, opened 2019-08-24
fix: set menu items spacing to equal

Nayananga:


zregvart:
@Nayananga I think we adjust the `padding-right` to `1.5rem` for the `navbar-item` to get the same effect here:

https://github.com/apache/camel-website/blob/06ff2e55d43df4d654b96be2da39354ddb8dd076/antora-ui-camel/src/css/header.css#L130-L133

antora-ui-camel/src/css/header.css
```diff
@@ -130,6 +130,8 @@ body {
 .navbar-item {
   flex: none;
   margin-bottom: 1px;
+  font-weight: bolder;
```
Nayananga: this made menu items in burger menu more visible

zregvart: This is overridden here:

https://github.com/apache/camel-website/blob/fb0ff7bbfe19b2b98d7be22d158a3676c55dc85d/antora-ui-camel/src/css/nav.css#L187

So it doesn't have an effect I'm afraid. Also would be good to use variables so we can tweak from vars.css.

zregvart:
Perfect, thanks!

PR #85, opened 2019-09-08
CAMEL-13933: Add menu item for EIP patterns

alexkieling:


.gitignore
```diff
@@ -3,6 +3,7 @@ target
 node_modules
 public
 documentation
+/static/
```
oscerd: Is this correct?

alexkieling: My bad. For some reason, I had lots of files under /static and thought they had been generated by some build script. I've removed it.

PR #94, opened 2019-10-17
Allow doc content to fill full available width

JoshBhurruthMO:
Currently, for viewports >1024px the doc class limits max-width to 46rem. This cuts off the final columns  for Query Parameter tables within Component docs. This limitation doesn't appear to be required

### DIFF
--- a/antora-ui-camel/src/css/vars.css
+++ b/antora-ui-camel/src/css/vars.css
@@ -131,7 +131,7 @@
   --nav-panel-height--desktop: calc(var(--nav-height--desktop) - var(--drawer-height));
   --nav-width: calc(270 / var(--rem-base) * 1rem);
   --doc-max-width: calc(720 / var(--rem-base) * 1rem);
-  --doc-max-width--desktop: calc(828 / var(--rem-base) * 1rem);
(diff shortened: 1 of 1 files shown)


zregvart:
Thank you for your contribution :100: 

The research in readability suggests having a 50-60, or a maximum of 75 characters per line. Here's an [article](https://baymard.com/blog/line-length-readability) on that. I guess that's where the motivation comes from.

We have a [preview](https://builds.apache.org/view/C/view/Apache%20Camel/job/Camel.website/view/change-requests/job/PR-94/Preview/) as a part of the PR checks, if anyone wishes to take a look at and see for themselves.

bedlaj:
I personally like this much more. Current width is IMHO too low for larger tables. Just my two cents, I am not an expert in typography.

zregvart:
I wonder if it would make sense to add way to toggle column/wide mode of the documentation. Like a link at the top of the documentation that switches the mode.

zregvart:
Alternate proposal in #107 to add a toggle for Wide / Narrow text, please have a look.

zregvart:
We have merged #107 as an alternative. Thanks @JoshBhurruthMO :)

PR #107, opened 2019-11-04
chore: add width toggle

zregvart:
This adds a toggle in the toolbar for documentation built by Antora to
switch between narrow (column) mode and wide (full width).

Perhaps as an alternative to #94 by @JoshBhurruthMO we can have the full width as an option.

### DIFF
--- a/antora-ui-camel/src/css/site.css
+++ b/antora-ui-camel/src/css/site.css
@@ -5,6 +5,7 @@
 @import "body.css";
 @import "nav.css";
 @import "main.css";
+@import "toggle-width.css";
 @import "toolbar.css";
 @import "breadcrumbs.css";
 @import "page-versions.css";
--- a/antora-ui-camel/src/css/toggle-width.css
+++ b/antora-ui-camel/src/css/toggle-width.css
@@ -1,0 +1,23 @@
+#toggle-width {
+  display: none;
+}
+
+#toggle-width:checked + article {
+  max-width: none;
+}
+
+#toggle-width-label {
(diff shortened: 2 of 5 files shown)


zregvart:
This adds the "Wide / Narrow" toggle, I wasn't sure if ARIA labels are needed. I guess they probably are, or the label should be marked as ignored

Screenshot:

Screenshot_2019-11-04 ActiveMQ Component - Apache Camel

Preview link:

https://builds.apache.org/job/Camel.website/job/PR-107/Preview/components/latest/activemq-component.html

djencks:
If the goal here is to provide a temporary way to view tables that are extra-wide, hiding the nav on the left (so it's a drop-down like on narrow screens) might also be useful.  I didn't look at the other proposal, but making this width change toggle-able is a very good idea.... the wider text is definitely harder to read than the default, but the effect is subtle enough that someone might not notice the strain.

zregvart:
@djencks thanks for your feedback. This would change only the text width (have a look at the preview), the navigation would stay as it is.

djencks:
Yes, I did look at your preview.  I’m suggesting that you consider also hiding the navigation:

— even wider tables would be easily visible
— regular text would look very odd, so people would be more apt to toggle back to narrow to ease the more obvious strain, rather than staying in wide-mode where, with your proposal, the strain might be vaguely unpleasant but not enough to easily notice.
— The page would be more symmetrical side-to-side.

BTW I didn't easily find a table that wasn't completely visible in the narrow view.

PR #113, opened 2019-12-02
Added missing camel-blueprint xsd after 3.0.0 release

oscerd:
@zregvart @davsclaus 

I added the xsd taken from camel-blueprint jars on maven central and by renaming it. I think this is what we were doing through svn more or less.


oscerd:
I'll do the same for spring and the other missing bits.

PR #120, opened 2019-12-04
Some hack to put Camel 3 is here on front page.

davsclaus:


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-120/Preview/index.html), since it pushes the content down it looks a bit empty with lots of white space, how about something like:

Screenshot_2019-12-04 Home - Apache Camel(1)

davsclaus:
Yeah that is better, or we could set it instead of "Integration you want"

zregvart:
I've created the two options in #121 and #122, let's see the previews (I'll post the links) and decide what works best.

PR #129, opened 2019-12-16
Allow to embed youtube videos

apupier:
I don't know how to test if the config is correct. it is working without the configuration locally but not on the live website


oscerd:
Thanks

oscerd:
In the preview it was good, on the live site it doesn't work, at least for me. So maybe a link would be better

zregvart:
Preview is served by Jenkins, so `.htaccess` is not taken into account.

PR #132, opened 2019-12-17
Re-add blogpost about Language Support Tooling 0.0.20

apupier:
- policies allows to publish without linking to original blogpost (even
if not written publicly, it will be fixed soon)
- provide image with link to the youtube video

Signed-off-by: Aurélien Pupier <apupier@redhat.com>


apupier:
sorry for the multi-reviews

PR #156, opened 2020-02-12
Add camel-kafka-connector to projects list

omarsmak:
Per title


zregvart:
Thanks @omarsmak I forgot about that!

PR #162, opened 2020-02-26
Fix GSOC year in template

DenisIstomin:


zregvart:
Thanks @DenisIstomin!

PR #81, opened 2019-08-21
chore: add search

zregvart:
Work in progress, first pass at the Algolia search implementation. Very work in progress with very rough design at this point.

PR #163, opened 2020-02-26
Pr/algolia search

Nayananga:
I think we haven't fully indexed the website so antora search indexes are missing from the search results
this work is related to #81


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-163/Preview/). I think we could do with a bit of styling on the search results, currently it looks like this:

Screenshot_2020-02-26 https builds apache org

There's also an error that gets logged in the browser console:

```
TypeError: n._snippetResult is undefined
```

It's good to see that the V4 of the Algolia search JavaScript is only 4.8KB (previous was 100+KB). I don't think it would load on the website due to content security policy:

https://github.com/apache/camel-website/blob/d35d9e3be385a7a4bdf958a7b28d8a2bfaa15999/static/.htaccess#L1326

We need to serve the resources from the same server, that would mean that we need to build/bundle with the theme (IIRC).

I wonder why we don't get the component reference indexed, I think in the end the best outcome for us would be to generate the index ourselves, but this is a good start.

djencks:
I’m just learning about search in Antora…. is the camel site too big for lunr search?

thanks
David Jencks

Nayananga:
@zregvart I`ll work on the styling and the console error.
>I wonder why we don't get the component reference indexed

If I remember correctly I showed you an algolia search demo with full site indexing. Perhaps this may be because of a configuration issue in that algolia crawler?  


>I think in the end the best outcome for us would be to generate the index ourselves,

+1

zregvart:
> I’m just learning about search in Antora…. is the camel site too big for lunr search?

@djencks we render to 4075 HTML documents. I saw a comment on Antora issue tracker that lunr is not a good choice for larger sites, not sure if 4075 is too large or not but Algolia seems to be used widely.

djencks:
Naively generating a lunr index on the antora site using antor-lunr, the search-index.js is about 75 mb, so the site is probably too large for lunr.

Nayananga:
@zregvart Hello Mr Zoran, I made a few changes I would like to have more suggestions.Thanks

antora-ui-camel/src/partials/footer-scripts.hbs
```diff
@@ -1,5 +1,6 @@
 <script src="{{uiRootPath}}{{asset 'js/site.js'}}"></script>
 <script async src="{{uiRootPath}}{{asset 'js/vendor/highlight.js'}}"></script>
 <script async src="{{uiRootPath}}{{asset 'js/vendor/svg4everybody.js'}}"></script>
+<script async src="https://cdn.jsdelivr.net/npm/algoliasearch@4/dist/algoliasearch-lite.umd.js"></script>
```
zregvart: The content security policy won't load this on the production site. We need to put this script in the [`vendor`](https://github.com/apache/camel-website/tree/master/antora-ui-camel/src/js/vendor) directory and load it from there. Just like the two other scripts we have here.

layouts/partials/footer.html
```diff
@@ -41,6 +41,7 @@
     <script src="{{ path.Join "_" (index .Site.Data "rev-manifest" "js/site.js") | relURL }}"></script>
     <script src="{{ path.Join "_" (index .Site.Data "rev-manifest" "js/vendor/highlight.js") | relURL }}"></script>
     <script src="{{ path.Join "_" (index .Site.Data "rev-manifest" "js/vendor/svg4everybody.js") | relURL }}"></script>
+    <script async src="https://cdn.jsdelivr.net/npm/algoliasearch@4/dist/algoliasearch-lite.umd.js"></script>
```
zregvart: We also need to load from the `vendor` directory here.

zregvart:
I think we just need to vendor the algolia JavaScript file to have it load in production. We can further refine the design upon the feedback we get for the search functionality.

antora-ui-camel/src/partials/footer-scripts.hbs
Nayananga: @zregvart I tried to solve this by 
first
>yarn add algoliasearch in antora-ui-camel 

then
>added a new algolia_search.js file into vendor folder which has
;(function () {
  var algoliasearch = require('algoliasearch')
  algoliassearch()
})()

finally
>add that script into to footer scrips as you mentioned

then I`m having these errors, any thoughts? 

Screenshot from 2020-03-03 08-06-46
Screenshot from 2020-03-03 08-05-37

djencks:
You need to have the file name be aloglia_search.bundle.js so that the UI builder will run browserify on the contents and convert the require into an “include”

Hope this helps :-)

David Jencks

antora-ui-camel/src/partials/footer-scripts.hbs
zregvart: You need to copy the file to the vendor directory, the Antora UI build is not using the dependencies declared in package.json. The idea of vendoring dependencies is that we copy the dependency that we use into our own source tree and not pull it in through any dependency management (such as npm or yarn).

djencks: It's much simpler to use the ui builder to pull in and browserify the sources. You do this with code like Nayananga suggested:
```
;(function () {
var algoliasearch = require('algoliasearch')
return algoliasearch()
})()
````
listing the `algoliasearch` in your package.json for the ui project, but naming the file `algoliasearch.bundle.js`.  The 'bundle' in the filename tells the builder to convert the require into a sort of 'include'.

Nayananga: @djencks Thanks for the suggestion, I was able to go half way through in that way by renaming the script, but then it gave me an error
> window.algoliasearch is not a function.

So I had to copy 06-search.js into the algoliasearch.bundle.js (which I`ve commited just now) now its work fine :)

zregvart:
The preview is [here](https://builds.apache.org/job/Camel.website/view/change-requests/job/PR-163/Preview/index.html).

zregvart:
I think we should merge this, it's not perfect but it works. The two outstanding issues are the design polish and having our own data built for the index. The current Algolia index is based on Algolia crawling the website and it doesn't seem to include some pages and it is not 100% accurate.

PR #167, opened 2020-03-06
Fixed right/left margin issue

AghaSaad04:

## Issue:
Article containers right margin is larger than the left margin

image

## Solution:
The issue was with the max-width property of the "doc" class. It was limiting the content to 46em max-width. I have removed the max-width property and now the result is:

image

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -3,7 +3,6 @@
   font-size: var(--doc-font-size);
   line-height: var(--doc-line-height);
   margin: var(--doc-margin);
-  max-width: var(--doc-max-width);
   padding: 0 1rem 4rem;
 }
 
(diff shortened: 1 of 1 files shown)


AghaSaad04:
I am thinking of removing these variables from the vars.css file as well. Should I do that?
image

zregvart:
@AghaSaad04 thank you for your contribution.

We have the "Wide / Narrow" link at the top right of the page that toggles between the wide or narrow (default) widths of text. We discussed this before on #94 and #107. The solution you propose breaks the "Wide / Narrow" functionality.

I'm still uncertain if we should have the content centered or offer the wider view of the text by default. There's a comment on [the issue](https://issues.apache.org/jira/browse/CAMEL-13881?focusedCommentId=17053138&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17053138) about this.

I would appreciate some research on this topic with regards to readability and usability to help reach a decision.

AghaSaad04:
Sure, thanks @zregvart. I will look more into this. :)

PR #170, opened 2020-03-06
Searched result styling

AghaSaad04:

## Issues:
- Search results push the website's Navbar.
- Searched result div should be scrollable.

## Solution:
- I have fixed the overlapping issue of the navbar items and searched result fiv. I have also added scrollbar to searched result div.

## Before:
image

## After:
image

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -3,6 +3,7 @@
   font-size: var(--doc-font-size);
   line-height: var(--doc-line-height);
   margin: var(--doc-margin);
+  max-width: var(--doc-max-width);
   padding: 0 1rem 4rem;
 }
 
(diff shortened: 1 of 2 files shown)


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-170/Preview/index.html).

Nayananga:
@AghaSaad04 this is great, I didn't know adding scroll bar is possible with CSS.

zregvart:
I think we can merge this. Looking at other implementations of Algolia search UI, like the one on https://gohugo.io/, https://www.gatsbyjs.org/ or https://reactjs.org/ they do not feature scrollbars. What is done there is that the search results are limited to 5 hits (we have 10) and the presentation is tailored to that: the results are presented in wider form and there's more distinction between each result.

This is an improvement on what we currently have, I think we can do better than this.

PR #173, opened 2020-03-07
Changed "what's new" to "What's new?

psingh587:
"what's new" seemed a little ambiguous to me, I think "What's new? " is looks better in its place. 
P.S - this is my first ever pull request. just happy to even make it.


zregvart:
Thanks!

PR #175, opened 2020-03-07
Replace table with unordered list

AghaSaad04:

Index.md file was using table structure with inconsistent strong properties. That was making the page less attractive.
I have replaced the table structure with the unordered list and removed the inconsistency.

### Before:
image

### After:

image


zregvart:
Thanks!

PR #176, opened 2020-03-08
Modify README file to install node using nvm

AemieJ:
`nvm default alias 10` didn't work on my system as expected. Hence, I included the method `nvm use 10.0.0` to switch to node version 10 in your local system. Also, I have included a step to step guide for install nvm on the system for the ease of people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,18 @@ Node on your machine.
 - Install using [Homebrew](https://brew.sh/) and [Node formulae](https://formulae.brew.sh/formula/node)
 - Install from official [Node packages](https://nodejs.org/en/download/)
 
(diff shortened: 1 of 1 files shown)


AemieJ:
The only reason I changed it to use `nvm use <version>` is 'cause it will work on each system.

oscerd:
What was the problem you had? What are "all the systems"?

AemieJ:
`nvm use <version>` will

zregvart:
I also run `nvm use v10`, I guess we do need to update the README I didn't notice it was out of date.

AemieJ:
> I also run `nvm use v10`, I guess we do need to update the README I didn't notice it was out of date.

I will make the changes if required by you. Also, I had a request to make that the glowing effect on the front page around **Get Started** doesn't look professional.

PR #177, opened 2020-03-08
Update README to include installation guide on nvm 

AemieJ:
A step by step guide added to install nvm on the system for the ease of users and revert back to `nvm alias default 10`.


zregvart:
Thanks!

PR #178, opened 2020-03-08
Made table style to be stripped and added padding

dzekem:
Added styling to the table by making the table header to have the camel orange background color and made the table to be striped as shown on the screen shot
Screenshot from 2020-03-08 14-12-18

### DIFF
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -133,4 +133,28 @@ article.blog p {
   .blog.list {
     flex-direction: row;
   }
+
+  table {
+    width: 100%;
+    border-collapse: collapse;
+  }
+
+  th,
+  td {
+    padding: 0.875rem;
+  }
+
+  table,
+  th,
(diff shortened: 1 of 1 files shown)


oscerd:
ok to test

zregvart:
Thanks, it's much better this way! For my taste I would go without table borders as we don't use them in other uses of tables in the documentation.

zregvart:
Perhaps I would mute the colors a bit as well. Here's a quick hack in the CSS inspector:

Screenshot_2020-03-09 Apache Camel 2020 Release Schedule - Apache Camel

That makes it a bit apparent that we also need to center align the `td` content.

zregvart:
And I just noticed, we can't make globally scoped rules in CSS. For example the `table` rule added here applies to all tables on the website. We have slightly different design for tables in the documentation. All rules need to be scoped to the `.blog` CSS class. That is instead of the selector `table` the selector needs to be `.blog table`.

PR #179, opened 2020-03-08
Fixed articles left and right margins

dzekem:
Fixed article left and right margins by setting them to **auto** since the max-width attribute was defined, this aligns the content of the article to the center.

Screenshot from 2020-03-08 15-58-23

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -2,7 +2,8 @@
   color: var(--doc-font-color);
   font-size: var(--doc-font-size);
   line-height: var(--doc-line-height);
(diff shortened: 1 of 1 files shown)


zregvart:
Thank you, I think this is better than what we considered before. It doesn't break the `Wide / Narrow` functionality we have, so those folk that wish to read the text at full width still have the option to do so.

zregvart:
Here's the [preview](https://builds.apache.org/job/Camel.website/job/PR-179/Preview/manual/latest/book-getting-started.html).

zregvart:
I think we can merge this now, thanks @dzekem!

dzekem:
Thank you all for the approvals.

PR #185, opened 2020-03-10
changed display:none to visible for swoosh animation to appear on mob…

CSElonewolf:
Before Change:-
beforechange

After Change :-
changed

This looks good as I have seen such changes exists in website like [swoosh animation in mobile](https://www.gssoc.tech/index.html#about)

gssoc

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -232,7 +232,7 @@ section.frontpage h1 {
   }
 
   header.frontpage svg {
-    display: none;
+    display: visible;
   }
 
   .box {
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -11,7 +11,7 @@ header.frontpage h1 {
   font-size: 5rem;
   font-weight: bolder;
   color: var(--heading-font-color);
-  margin-top: 5rem;
+  margin-top: 3rem;
   text-align: center;
   text-transform: none;
   text-shadow: -1px -1px 1px #fff5, 1px -1px 1px #fff5, -1px 1px 1px #fff5, 1px 1px 1px #fff5;
@@ -232,7 +232,7 @@ section.frontpage h1 {
   }
 
   header.frontpage svg {
-    height: 12rem;
+    height: 10rem;
   }
 
   .box {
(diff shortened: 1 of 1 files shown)


zregvart:
I think the swoosh looks a bit awkwardly shrunk when compressed from the side. We might need to resize in the vertical space as well, perhaps by reducing the font sizes on smaller resolutions.

For example, this looks a bit better to me:
Screen Shot 2020-03-10 at 11 52 43

zregvart:
Also one note, do we know about battery usage of animations on mobile, I would appreciate that we don't end up wasting folks power.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -232,7 +232,7 @@ section.frontpage h1 {
   }
 
   header.frontpage svg {
-    display: none;
+    display: visible;
```
zregvart: Perhaps removing the whole rule would serve the same purpose?

CSElonewolf:
> Also one note, do we know about battery usage of animations on mobile, I would appreciate that we don't end up wasting folks power.

I guess mostly the site is visited either in Desktop or Laptop and seldom on mobile.So, battery is not an issue.

antora-ui-camel/src/css/frontpage.css
CSElonewolf: Ok.Sure.Thats a great Idea.I was too thinking about the abnormal font-size of the <h1> tag content as soon as the screen was reduced from 1023px.
I will be changing the font-size and will also increase the opacity of the shadow of the <h1> tag's content.

zregvart:
I would reduce the top margin of the `header.frontpage h1` and reduce the height of the swoosh. So that the vertical distance between the (1) and (2) is symmetrical and that the thickest part of the swoosh (3) is not at such great contrast to the thinnest part of the swoosh.

Screenshot_2020-03-10 Home - Apache Camel

zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-185/Preview/index.html).

CSElonewolf:
Annotation 2020-03-11 102627

This is the Preview

zregvart:
Thanks!

CSElonewolf:
Most welcome @zregvart

PR #190, opened 2020-03-11
CAMEL-14693 : Improved the frontpage header text size

psingh587:
Regarding the issue raised by me 
https://issues.apache.org/jira/browse/CAMEL-14693

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -228,7 +228,7 @@ section.frontpage h1 {
 
 @media screen and (max-width: 1023px) {
   header.frontpage h1 {
-    font-size: 15vw;
+    font-size: 9vw;
   }
 
   header.frontpage svg {


psingh587:
WhatsApp Image 2020-03-11 at 12 51 41

After this changes it looks something like this

psingh587:
@zregvart  
Sorry sir, looks like some checks have failed.
Please guide me where i was wrong.

zregvart:
@psingh587 we build in containers and sometimes we have issues with stopping them (unsure of the reason why), I've restarted the build, it's usually on the second go that the issue doesn't manifest.

zregvart:
I think we have this issue fixed in #185 already.

psingh587:
> I think we have this issue fixed in #185 already.

Sir , i had checked that PR. But i guess its a little different from what i have tried to do here.
I have decreased the font-size when screen width turned to 1023 px which i think stays in PR #185. I thought it would be good if the font is not that big that it is now when screen width is 1023px.

psingh587:
> I think we have this issue fixed in #185 already.

Yes sir. Just checked it :)
I will close this PR then

zregvart:
The preview is [here](https://builds.apache.org/job/Camel.website/job/PR-190/Preview/index.html). @psingh587 if you'd like to rebase on the current `master` branch and see if there are more improvements you can add on top of #185 you're more than welcome.

PR #191, opened 2020-03-11
solved search box interferes with menu and logo on smaller resolutions

CSElonewolf:
[CAMEL-14694](https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-14694?filter=allopenissues)

Before:-
image 3

After:-
image2

Updated the margin-top ,padding and top of .navbar-search input and #search_results respectively

PR #192, opened 2020-03-11
Made the logo shorter in shorter resolutions of screen

psingh587:
WhatsApp Video 2020-03-11 at 18 41 48

After the change.

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -228,7 +228,7 @@ section.frontpage h1 {
 
 @media screen and (max-width: 1023px) {
   header.frontpage h1 {
-    font-size: 15vw;
+    font-size: 9vw;
   }
 
   header.frontpage svg {
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -191,6 +191,22 @@ body {
   }
 }
 
+@media screen and (max-width: 607px) {
+  .nav-logo {
+    background-image: url('../img/logo-d.svg');
+    background-repeat: no-repeat;
+    background-size: 3.5rem;
+    background-position-x: 7px;
+    background-position-y: 3.7px;
+    width: 9rem;
+    position: relative;
+  }
+
+  .nav-logo span {
+    display: none;
+  }
+}
+
 @media screen and (min-width: 1024px) {
   .navbar,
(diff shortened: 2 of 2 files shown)


psingh587:
@zregvart 
Here's my take on it, Sir.
The bigger logo which is there right now gets hidden behind the search bar, so i tried to making the logo itself smaller. We can keep both the logo(smaller one ) and search box together for smaller resolutions this way ( ~ 471px being the smallest after which the logo starts to hide, in that case we can try something else ).

CSElonewolf:
This is better idea @psingh587 .
@zregvart and @oscerd please have a look and let me know whether I should proceed with the issue #191 or not.

zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-192/Preview/index.html).

Nice work @psingh587, I think this is much better than what we have. I'm worried a bit about even smaller widths, I usually test with 360 and 411 widths in Firefox this profile is called "Galaxy S9/S9+ Android 7.0" and "Pixel 2 Android 8.0". It's not as nice with those.

I'm not sure what is the smallest width that we should support, so it might be fine regardless.

My opinion, and I don't have to be right, is that perhaps a combined approach with ideas from both @CSElonewolf and @psingh587 could be the winner here. Idea of putting the search in the menu is very interesting to me. In the long run I'd like the menu on smaller screens to take up the full width of the screen, that way we can increase the spacing and have larger touch targets.

psingh587:
@zregvart 
Thank you sir for your feedback.
I myself think that the idea of putting the search bar in the menu is quite nice. We can do that for resolutions below 471. But beyond 471 and till 1023 px widths i think this looks good with a smaller logo and search bar together

CSElonewolf:
Screenshot_2020-03-11-20-12-31-108_com android chrome


This is how the preview looks in my smartphone. :( issue remains.
But not to worry.Well then let me go ahead with modifying my own #191 . @zregvart 
Hopefully will submit a new commit today itself.

psingh587:
> Screenshot_2020-03-11-20-12-31-108_com android chrome
> 
> This is how the preview looks in my smartphone. :( issue remains.
> But not to worry.Well then let me go ahead with modifying my own #191 . @zregvart
> Hopefully will submit a new commit today itself.

Yes.. I have not made the changes for smaller screen width below 471. I think you can go with putting search bar in menu for it

PR #194, opened 2020-03-11
Mobile Menu Enhancement

avgupt:
Fixes: https://issues.apache.org/jira/browse/CAMEL-14695?filter=12348381

PR #196, opened 2020-03-11
Improvement upon the navbar-menu

psingh587:
This is regarding the merged PR #194 
  I have tried making some amendments to it 
@zregvart  @oscerd  Please check it.

This is how the navbar-menu covers the whole space on 1023px. What I feel is that it is not neccessary to have it cover the entire window here. 
1023-orginal

this is my version (when width is 1023px)
1023-myversion

and also i think in smaller resolutions it can cover the entire screen so here's my take on it..
Mob-local

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -170,6 +170,20 @@ body {
     flex-flow: column;
   }
 
+  .navbar-item.navbar-topics,
+  .navbar-link.navbar-topics {
+    color: var(--navbar-font-color);
+  }
+
+  .navbar-item {
+    font-weight: bold;
+  }
+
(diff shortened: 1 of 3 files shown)


psingh587:
Sir, is this okay.? if not how shall i proceed about this?
@zregvart @oscerd

psingh587:
the second commit does something like this 
WhatsApp Image 2020-03-12 at 05 53 56


  --the color is same as what the navbar items in bigger resolution have, also the dropdown list items have underline hover property .

psingh587:
@zregvart @oscerd  
Sir, How were these changes I made? 
How should I go about improving them if they aren't satisfactory?
Please guide

zregvart:
Thanks!

PR #197, opened 2020-03-12
fixed blog page categories width

jyotiattri:
Hi,
I have found an issue on the https://camel.apache.org/blog/ for mobile screens. The main menu is going outside the page and there is a blank space in the bottom. Please check the following screenshot.
blog-page

This PR fixes this problem. Here is an updated screenshot after the fix.
changed-blog-page

@zregvart please check.

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -634,7 +634,7 @@ kbd,
 .stretch {
   width: 100%;
 }
-
+/* for mobile screens */
 @media screen and (max-width: 600px) {
   .doc aside {
     width: auto;
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -634,3 +634,9 @@ kbd,
 .stretch {
   width: 100%;
 }
+
+@media screen and (max-width: 600px) {
+  .doc aside {
+    width: auto;
+  }
+}
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -635,8 +635,8 @@ kbd,
   width: 100%;
 }
 /* for mobile screens */
-@media screen and (max-width: 600px) {
+@media screen and (max-width: 1023px) {
   .doc aside {
-    width: auto;
+    display: none;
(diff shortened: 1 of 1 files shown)


jyotiattri:
Hi @oscerd 
Is there anything that I can do fix the above error?

oscerd:
I started the build again

oscerd:
cc @zregvart

zregvart:
When we introduce new media selectors we take on the work to make sure that different parts of the website adhere to those selectors, this makes it difficult for us to maintain. That's why I'm hesitant to introduce new media selectors.

Perhaps this can be done within the existing media selector here:
https://github.com/apache/camel-website/blob/ce074b7f117927e3bfcedd785f10e6053bab058f/antora-ui-camel/src/css/doc.css#L39-L43

And perhaps on widths less than 769px we hide the categories menu altogether. The way we have it right now is that it's pushed to the bottom of the page which is not as nice and makes folk scroll to the bottom to see it.

jyotiattri:
For mobile screens, this media query is not working because it works only more than 769px or equal to 769px width.
For hiding the categories menu on smaller screens we should need a media query for a small screen like header.css used for mobile menus.
@zregvart Please help me on this issue.

antora-ui-camel/src/css/doc.css
```diff
@@ -634,3 +634,9 @@ kbd,
 .stretch {
   width: 100%;
 }
+/* for mobile screens */
+@media screen and (max-width: 600px) {
+  .doc aside {
+    width: auto;
+  }
+}
```
zregvart: Perhaps we could remove the categories on any screen smaller than 1024px, like this:

```suggestion
@media screen and (max-width:1023px) {
  .doc aside {
    display: none;
  }
}
```

That way we don't introduce any new media breakpoints.

zregvart:
Thanks!

zregvart:
I had to revert these changes, the `doc.css` is reserved for styling the documentation, these changes need to go to either `blog.css` or to `category.css` and they need to be scoped via css rule i.e. prefixed with `.blog`. Not 100% sure on the file, on first thought `blog.css` is more appropriate.

PR #199, opened 2020-03-12
Removed unused html from header

AghaSaad04:

I have removed the unused HTML from the header. It was not visible to users because of the styling and was of no use.

### DIFF
--- a/antora-ui-camel/src/css/nav.css
+++ b/antora-ui-camel/src/css/nav.css
@@ -294,14 +294,3 @@ html.is-clipped--nav {
   background-position-y: 10px;
   width: 11rem;
 }
-
-.nav-logo span {
-  position: absolute;
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px;
-  overflow: hidden;
-  clip: rect(0, 0, 0, 0);
-  border: 0;
-}
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -33,7 +33,7 @@
     <header class="header">
         <nav class="navbar">
             <div class="navbar-brand">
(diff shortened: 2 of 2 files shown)


zregvart:
I think the idea of that non visible <span> is to help folk that don't load images or use a screen reader to have the text "Apache Camel" there for them.
We do have a `title` attribute on the logo, so that might help. Would appreciate some research on this.

If we merge this and remove the span this section of the CSS also needs to be removed:

https://github.com/apache/camel-website/blob/9c2ddbe4130d230f6466c88ff1d5276f952c1d42/antora-ui-camel/src/css/nav.css#L298-L307

AghaSaad04:
Sure @zregvart. I will look more into this and will update the PR if required.

AghaSaad04:
@zregvart I looked more into this and to my best knowledge we may have to remove the CSS.

zregvart:
> @zregvart I looked more into this and to my best knowledge we may have to remove the CSS.

What about screen readers and folk that don't load images, can we help them also?

AghaSaad04:
@zregvart This is what I have got after researching about the title attribute:

> Values of the title attribute may be rendered by user agents in a variety of ways. For instance, visual browsers frequently display the title as a "tool tip" (a short message that appears when the pointing device pauses over an object). **Audio user agents** may speak the title information in a similar context. For example, setting the attribute on a link allows user agents (visual and non-visual) to tell users about the nature of the linked resource

Anyone interested can visit [this](https://stackoverflow.com/questions/4216035/css-background-image-alt-attribute)

So I think we are helping the screen readers and all those dont load images.

AghaSaad04:
Thanks, @zregvart for the article you mentioned, I will read that now :)

zregvart:
@AghaSaad04 when we change the templates in Hugo layout we also need to make the corresponding change to the Antora UI. See a58201e.

AghaSaad04:
@zregvart thanks for the info. I will update both from now on if I make any change :+1:

PR #200, opened 2020-03-12
fixed issue for content wider than screen for pages

jyotiattri:
This adds a fix for  **Content wider than screen (16 pages)**  from [Mobile usability issues reported in Google search console]( https://issues.apache.org/jira/browse/CAMEL-13800)

cc : @zregvart

### DIFF
--- a/antora-ui-camel/src/css/base.css
+++ b/antora-ui-camel/src/css/base.css
@@ -48,6 +48,7 @@ kbd,
 pre {
   color: var(--code-font-color);
   font-family: var(--monospace-font-family);
+  overflow-wrap: break-word;
 }
 
 b,
@@ -84,12 +85,4 @@ mark {
 
 .text-center {
   text-align: center;
-}
-
-@media screen (max-width: 1023px) and (min-width: 480px) {
-  code,
-  kbd,
-  pre {
-    overflow-wrap: break-word;
-  }
 }
--- a/antora-ui-camel/src/css/base.css
+++ b/antora-ui-camel/src/css/base.css
@@ -86,18 +86,10 @@ mark {
   text-align: center;
 }
 
-@media screen and (min-width: 769px) {
+@media screen (max-width: 1023px) and (min-width: 480px) {
   code,
   kbd,
   pre {
     overflow-wrap: break-word;
-  }
-}
-
-@media screen and (max-width: 626px) {
-  code,
-  kbd,
-  pre {
-    font-size: 3.5vw;
   }
 }
--- a/antora-ui-camel/src/css/base.css
+++ b/antora-ui-camel/src/css/base.css
@@ -85,3 +85,19 @@ mark {
 .text-center {
   text-align: center;
 }
+
+@media screen and (min-width: 769px) {
+  code,
+  kbd,
+  pre {
+    overflow-wrap: break-word;
+  }
+}
+
+@media screen and (max-width: 626px) {
+  code,
+  kbd,
+  pre {
+    font-size: 3.5vw;
+  }
+}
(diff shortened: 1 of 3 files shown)


jyotiattri:
Ok, I got it. I can change it.

antora-ui-camel/src/css/doc.css
```diff
@@ -40,6 +40,16 @@
   .doc > h1.page:first-child {
     margin-top: 2.5rem;
   }
+
+  .release dl {
+    overflow: scroll;
+  }
```
zregvart: This needs to go to [antora-ui-camel/src/css/release.css](https://github.com/apache/camel-website/blob/master/antora-ui-camel/src/css/release.css)

zregvart:
I'm not sure about using `vh` `font-size` unit inconsistently, at 626px this leads to quite a big difference between font sizes:

Screenshot_2020-03-15 ActiveMQ Component - Apache Camel

I think it's best that we focus on `overflow-wrap` and `overflow` changes here and on changing font-sizes in a separate issue/pull request.

jyotiattri:
Requested Changes are done in this PR.

jyotiattri:
Hi, @zregvart 
Please let me know is there something to do in this PR.

antora-ui-camel/src/css/base.css
```diff
@@ -83,3 +83,11 @@ mark {
 .text-center {
   text-align: center;
 }
+
+@media screen (max-width: 1023px) and (min-width: 480px) {
+  code,
+  kbd,
+  pre {
+    overflow-wrap: break-word;
```
zregvart: Does it make sense to remove the media query and put this in the existing rule regardless of the screen size?

antora-ui-camel/src/css/doc.css
```diff
@@ -634,3 +636,22 @@ kbd,
...
+@media screen and (max-width: 1023px) and (min-width: 480px) {
+  .doc > h1.page:first-child,
+  .static > h1:first-child {
+    font-size: inherit;
+  }
+
+  .doc h2:not(.discrete) {
+    overflow-wrap: break-word;
+  }
+}
```
zregvart: I think this needs to be rethought: these are some very targeted rules, I'd prefer if we came up with a principle rather than a fix for specific instances of text. This will be very difficult to maintain with the text changing.

antora-ui-camel/src/css/release.css
```diff
@@ -11,3 +11,9 @@
   grid-column-start: 2;
   margin-left: 1rem;
 }
+
+@media screen and (max-width: 1023px) and (min-width: 480px) {
+  .release dl {
+    overflow: scroll;
+  }
+}
```
zregvart: Why use a media query, could we put this in `.release dl` instead? What would be the downside of that?

antora-ui-camel/src/css/base.css
jyotiattri: Yes, it will work.

antora-ui-camel/src/css/release.css
jyotiattri: No, the downside of this. It will work fine without using the media query.

antora-ui-camel/src/css/doc.css
jyotiattri: In this, we have a different font size so I use this media query for some targeted areas font size.
I made changes in this and sent it again.

zregvart:
I've rebased this on current master branch, I'm interested to see if there are any issues with this with other changes that have happened in the meantime. So let's wait for the new preview.

jyotiattri:
@zregvart, Ok Thanks

zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-200/Preview/index.html).

zregvart:
Thanks!

PR #201, opened 2020-03-12
Fixed apache logo in mobile screen.

CSElonewolf:
This solves [CAMEL-14701](https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-14701?filter=allopenissues)
**Before**
beforechange

**After**
imagesss

Previously the `flex:50%` prevented to chnage the position but now `flex: 100%` in mobile screen does the job.


zregvart:
Thanks!

PR #205, opened 2020-03-12
Better Contrast Ratio and Accessibility

avgupt:
I have made some changes in the theme to increase the contrast ratio where it was needed.

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -51,7 +51,7 @@ header.frontpage a:active {
 .frontpage a.significant {
   margin: 1rem;
   padding: 0.8rem 1.5rem;
-  background: var(--color-camel-orange);
+  background: var(--color-camel-brown);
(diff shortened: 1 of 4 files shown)


aashnajena:
Hey @avgupt can you please put up some screenshots of the changes you have made? I don't want to be making the same changes again!

avgupt:
@zregvart I agree that the blue background is not the best choice, but the current contrast ratio value comes close to 2.8, but having blue on white gives a contrast ratio of 10.9.
Just an idea to change navbar text and background colours to something with better contrast ratio.

zregvart:
@avgupt there is a definite trade-off between design aesthetics and accessibility, this I feel swings to a place that is very much less favourable to design aesthetics. Can you try another approach? Perhaps research a bit for tools and recommendations first.

zregvart:
We haven't seen any new developments in 10 days so I think we can close this @avgupt feel free to reopen when you have further changes.

PR #219, opened 2020-03-14
CAMEL-14718: Improvize on the front-page of the website

AemieJ:
* In the front-page within the section where it states about Camel, Camel-K and provided with the button to read the documentation, it is observed that for a smaller screen width, the title and buttons are center-aligned but the paragraph isn't. Thus, I have made the paragraph center-aligned as well. This creates neatness in the front-page presentation.


zregvart:
Thanks!

PR #220, opened 2020-03-15
Improve the designing of the front-page

favourori:
First issue which was to align camel section etxt properly on smaller screens was fixed by

 section.frontpage.projects .project p {
      text-align: center;
 }
to the ...@media screen and (max-width: 1023px)...

Second issueswhich was to add a nice hover effect was fixed by add an after pseudo class to the links in order to achieve an hover
effect as was suggested here link creative button hover effects.


AemieJ:
I have already created a center alignment improvement to Frontpage and created a push down hover effect. If you don't mind, could I push the changes for the push down hover effect?

PR #221, opened 2020-03-15
CAMEL-14718: add the push down hover effect to the buttons in frontpage

AemieJ:
* The problem I observed was that in different screen width including the mobile version, when I hovered over the buttons, the glow of each button used to overlap the other. Hence, I removed the outer glow effect for the buttons.

* Instead, I created a push-down hover effect for the button which will look sophisticated on the frontpage.


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-221/Preview/index.html).

AemieJ:
I think it works. As in the glow effect, the spread of glow was overlapping the other buttons.

PR #223, opened 2020-03-16
Translate property to blog aside category list items on hover

psingh587:
Screenshot from 2020-03-16 16-08-09

Added a translate property , which will get activated by hovering upon this list items


psingh587:
How should i improve it? @zregvart @oscerd

PR #229, opened 2020-03-17
Fix side nav-bar issues

aashnajena:
In the current website, the left-side nav-bar has the following issues :
- on hover, some of the links don't get underlined : this is because the CSS comes from a:hover so only external links show the change on hover. I have changed this so that all menu items turn black and get underlined on hovering. Eg. the "Supported expression languages" and "Frequently asked questions" on the main menu of user manual don't show any hover effect.
- In the User Manual, the submenu list under Architecture and Integration patterns is so long that if you open the drop-down, the menu becomes very long and difficult to navigate (you can't see the other main-menu items due to the long sub-menu list). I have changed this such that if the submenu takes > 200px height, it becomes scrollable instead of taking up the entire space of the main-menu. I have styled the scrollbar to match the aesthetics.
- The drop-down arrow buttons on the left seem uneven, since they are not present for every item and hence the menu looks oddly indented. I have revamped the menu - please see the screenshot below. Now user can make out the difference between menu and content easily, each item of the menu is well separated, the drop-down buttons are on the right for even-ness. The sub-menus are scrollable.
- On using the quick search present in "components", extra border lines were appearing because only the nav-links were getting hidden, not the entire nav-item, so I changed that as well.

The problem :
sidebar-menu-problem

The menu design :
menu-design

Sub-menu design (scrollable) 
menu-scroll

### DIFF
--- a/antora-ui-camel/src/css/nav.css
+++ b/antora-ui-camel/src/css/nav.css
@@ -88,11 +88,13 @@ html.is-clipped--nav {
   left: 0;
 }
 
-.nav-panel-menu::-webkit-scrollbar {
+.nav-panel-menu::-webkit-scrollbar,
+.nav-list::-webkit-scrollbar {
   width: 0.25rem;
 }
 
-.nav-panel-menu::-webkit-scrollbar-thumb {
+.nav-panel-menu::-webkit-scrollbar-thumb,
+.nav-list::-webkit-scrollbar-thumb {
   background-color: var(--nav-border-color);
 }
 
@@ -145,6 +147,11 @@ html.is-clipped--nav {
   margin-bottom: 0.5rem;
 }
 
+.nav-list .nav-list .nav-list {
+  max-height: 200px;
+  overflow-y: auto;
+}
+
 .nav-item {
   list-style: none;
   margin-top: 0.5em;
@@ -184,6 +191,13 @@ html.is-clipped--nav {
 .is-current-page > .nav-link,
 .is-current-page > .nav-text {
   font-weight: var(--body-font-weight-bold);
+}
+
+.nav-link:hover,
+.nav-text:hover {
+  color: var(--color-camel-orange);
+  font-weight: var(--body-font-weight-bold);
+  text-decoration: underline;
 }
 
 .nav-panel-explore {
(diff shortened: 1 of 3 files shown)


aashnajena:
I'm sorry, I didn't understand the part about hover and menu reflow. If the user clicks on the drop-down arrow, the submenu opens up (and if sub-menu is too long, it becomes scrollable). If the grey background or the lines seem very distractive for you, I can remove that. Is that the concern?

zregvart:
@aashnajena this is what I see on the [preview](https://builds.apache.org/job/Camel.website/job/PR-229/Preview/manual/latest/):

Peek 2020-03-26 10-54

See how the bold text forces the reflow on hover making the text jump below it.

To me this seems way to busy and I personally prefer the simpler version we currently have.

aashnajena:
Ahh I see, I agree, this is not good. So should I undo this bold effect (so that menu does not reflow)? Or would you prefer if I close this pull request completely...

aashnajena:
I thought about your comment and you're right, the design is too busy. Does removing the bold effect and the delimiting lines make the design less busy for you? 

sidebar-menu

zregvart:
I don't think there's an issue with making the currently selected topic bold. This is set on load of the page and will not cause reflows when user interacts with the webpage.

I think the idea of putting the expand/contract triangle on the right is a bit troublesome, it's close to the scrollbars (by default on the right side).

There's also an issue with additional scrollbar appearing when expanding certain sections like here:

Screenshot_2020-03-27 Apache Camel user manual - Apache Camel

This is also poor experience for the user as the user now has to use two scrollbars, think about the issues with the mouse scroll wheel.

So going back to the problem statement:
- _on hover, some of the links don't get underlined_ - the elements that don't get underline effect are not in fact links, they only serve to expand contract sections for example in the "User manual" the "Domain Specific Languages" is not a link if you click on it it only expands topics within it so I don't see a problem with that
- _the submenu list under Architecture and Integration patterns is so long that if you open the drop-down_ - either way scrollbar appears, having smaller area to scroll is worse in my opinion (see above)
- _drop-down arrow buttons on the left seem uneven, since they are not present for every item and hence the menu looks oddly indented_ - not seeing the unevenness, they would be present only for topics that have topics nested within them, alignment looks correct to me, perhaps you can provide a screenshot of an example where this is a problem

We have a [separate issue](https://issues.apache.org/jira/browse/CAMEL-14797) for the discrepancy in the component quick search, please reach out to @psingh587 so there's no duplicate work for that.

aashnajena:
I understand the decisions behind those features now, thank you very much for explaining it to me. I think it's best if I close this PR and work on other issues!

PR #234, opened 2020-03-18
Added an anchor tag to allow the users to reach the top of the page

psingh587:
-For pages with large contents, once we scrolled all the way down at the bottom of the page, we had to scroll up again to reach the top of the page. So to make the part easier, i added anchor tag at the bottom of the page.
Screenshot from 2020-03-18 14-45-33


psingh587:
Hi @oscerd
i checked the shell script details. it showed : anchor not found from security/index.html to security/index.html#top
Could you please guide how to eradicate this error?

psingh587:
> This changes the layout in Hugo, we need a corresponding change in the Antora layouts as well. I don't see where the element with `id=top` is present so that's missing as well, without it this will not function.

Hi @zregvart  ,
about the id issue, actually I have tried this one in my local machine. It works perfectly. Upon clicking it, it takes to the top of the page without having an id="top". I wonder why, but it does work here.
Shall i make specific changes in antora layouts?

djencks:
Doesn't a link to #does-not-exist always take you to the top of the page?  Unless this is a standard technique in web design that  I don't know about, I think it would be better to find a solution that doesn't rely on what I'd term exception handling. For instance does <a href="#"... work?

psingh587:
> Doesn't a link to #does-not-exist always take you to the top of the page? Unless this is a standard technique in web design that I don't know about, I think it would be better to find a solution that doesn't rely on what I'd term exception handling. For instance does <a href="#"... work?

Hi @djencks 
Yes sir, I might have missed that part while learning. <a href="#"> this works too. I will try to come up with a better solution then. Thanks.

zregvart:
Today I learned about `#top` not needing to be defined in the markup. Thanks for that @psingh587 and @djencks. I've found a note in the examples [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) to attest to that.

@psingh587 we need the same change in the Antora UI layout here (just add the top of the page markup):

https://github.com/apache/camel-website/blob/3120c065b92dfca83059551d416a859134b89462/antora-ui-camel/src/partials/footer-content.hbs#L1

zregvart:
We also need to configure the [link-checker](https://github.com/timaschew/link-checker) to ignore links that target `#top`. I think we can add `--url-ignore='#top'` here:

https://github.com/apache/camel-website/blob/3120c065b92dfca83059551d416a859134b89462/package.json#L13

psingh587:
> @psingh587 we need the same change in the Antora UI layout here (just add the top of the page markup):
> 
> https://github.com/apache/camel-website/blob/3120c065b92dfca83059551d416a859134b89462/antora-ui-camel/src/partials/footer-content.hbs#L1

@zregvart 
 *This particular change comes under <div class="edit".. and not under <footer*
This is how the footer-content.hbs begins
` <footer>
        <div class="footer">
            <figure class="logo">
                Apache Camel Logo
            </figure>`

and this is how the footer.html begins.
`   

           <div class="edit"> <a title="Improve this document, receive free virtual hugs &hearts;" href="https://github.com/apache/camel-website/edit/master/content/{{ .File.Path | replaceRE "^/" "" }}">Edit this Page
            </a>
            <a href="#top" title="Reach the top of the page">Back to top</a>
    </div>
    <footer>
        <div class="footer">
            <figure class="logo">`


I guess there is no requirement to make any changes there. Please correct me if I'm wrong.

psingh587:
> We also need to configure the [link-checker](https://github.com/timaschew/link-checker) to ignore links that target `#top`. I think we can add `--url-ignore='#top'` here:
> 
> https://github.com/apache/camel-website/blob/3120c065b92dfca83059551d416a859134b89462/package.json#L13

Here in this, Do I have to add this particular line `--url-ignore='#top'` within the same lines as of link-checker like this?
 "check:links": "link-checker --disable-external --mkdocs --allow-hash-href public --url-ignore='#top'",

psingh587:
> We need the link to top of the page in the documentation as well and the link-checker needs to be configured.

@zregvart 
For files under documentation, every page is hard-coded. So I will have to change the each and every file under it to achieve this. Is this the only way to do this, or am I missing out on something?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "version": "1.0.0-SNAPSHOT",
-  "license": "Apache-2.0",
+  "license": "Apache-2.0",
```
zregvart: Nitpick: trailing whitespace

package.json
```diff
@@ -10,7 +10,7 @@
     "preview:antora": "antora --clean --fetch antora-playbook.yml",
     "preview:hugo": "hugo server -D",
     "check:xref": "antora --generator @antora/xref-validator antora-playbook.yml",
-    "check:links": "link-checker --disable-external --mkdocs --allow-hash-href public",
+    "check:links": "link-checker --disable-external --mkdocs --allow-hash-href public --url-ignore='#top'",
```
zregvart: :+1:

layouts/partials/footer.html
```diff
@@ -46,3 +51,4 @@
 
 </body>
 </html>
+
```
zregvart: Nitpick: empty line

antora-ui-camel/src/css/footer.css
```diff
@@ -71,7 +71,7 @@ footer .footer dl dd {
   margin-left: 80%;
 }
 
-@media screen and (max-width: 640px) {
+@media screen and (max-width: 1181px) {
```
psingh587: Screenshot from 2020-03-28 09-46-10

this I had to because, as can be seen, in the picture above, the blocks were stashing above each other. Now with this they don't and this particular thing applies for <div class="edit" only so it does not affect anything else.

psingh587:
@zregvart 
Is it okay to add a smooth scrolling effect when we click on Back To Top?
Rather than going to top at once, it would smoothly travel all its way up. In my opinion, this would be good. 

ezgif com-optimize

antora-ui-camel/src/css/footer.css
zregvart: There needs to be a better way, adding media breakpoints is not a good solution. With any change we do, say we increase/decrease the font size we need to test at every breakpoint, the total number of tests is layouts*breakpoints, so it makes it very difficult for us to test, that's why we can't be adding new breakpoints.

antora-playbook.yml
```diff
@@ -45,6 +45,9 @@ content:
     - url: git@github.com:apache/camel-spring-boot.git
       branches: master
       start_path: docs
+    - url: git@github.com:apache/camel-karaf.git
+      branches: master
+      start_path: docs
```
zregvart: This looks like an unrelated change, can you rebase to pick up the current playbook?

antora-ui-camel/src/css/footer.css
psingh587: So, Should I work in the previous breakpoint?

zregvart:
@psingh587 I have a concern about accessibility of smooth scrolling, read [this](https://css-tricks.com/smooth-scrolling-accessibility/).

psingh587:
> @psingh587 I have a concern about accessibility of smooth scrolling, read [this](https://css-tricks.com/smooth-scrolling-accessibility/).

@zregvart agreed with the fact that for longer pages, there would be a lot of content to be scrolled up at once and that could cause a dizzy effect for the user. Maybe, then I should drop this idea.

psingh587:
I have created a fresh PR #301 for this and it includes all the requested changes.

psingh587:
@zregvart 
Should I record this PR in Outreachy website or leave it?

zregvart:
@psingh587 please record this, you both raised the issue did the first implementation and reviewed the second one.

PR #238, opened 2020-03-18
Consistency in title color for blog page

rimshach:
I was going through the website and I found the title color different for blog's page as compared to the rest of the website.I think there should be consistency in the entire website so I changed the color

The color for titles or link in other pages:

image

The color for blog was :
image

I have changed it to :
image

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -116,7 +116,7 @@
 }
 
 .doc a {
-  color:#303284;
+  color: #303284;
 }
 
 .doc a:hover {
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -116,7 +116,7 @@
 }
 
 .doc a {
-  color: var(--link-font-color);
+  color:#303284;
 }
 
 .doc a:hover {


antora-ui-camel/src/css/doc.css
```diff
@@ -116,7 +116,7 @@
 }
 
 .doc a {
-  color: var(--link-font-color);
+  color: #303284;
```
zregvart: We use CSS variables, it's very difficult to maintain individual values scattered across CSS files.

rimshach: oh okay, there is an option under each heading link of "continue reading " ,that link and heading link redirects us to the same page, so would it be okay to remove the link from the heading ? 
Or as in the home page the link "What's New?" has been assigned color of a regular heading 
image

rimshach:
oh alright.I actually looked at the heading link of "what's new " at home page that was consistent with rest of the headings so I thought overall headings can have same color.
For this page  there is an option under each heading link of "continue reading " ,that link and heading link redirects us to the same page, so would it be okay to remove the link from the heading ?

antora-ui-camel/src/css/doc.css
rimshach: Oh , I'll consider this before making any further changes

zregvart:
@rimshach I think it makes sense to remove the heading link from the blog post list as we already have the "continue reading" link there.

PR #239, opened 2020-03-19
Issue/Fixed Quick lookup input box width

jyotiattri:
Quick lookup has a fixed width, it affects the menu text in mobile view and iPad view. Half of the menu text gets displayed after this input box which doesn't look good. It should have full width.

CC: @zregvart


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-239/Preview/components/latest/).

zregvart:
Thanks!

PR #247, opened 2020-03-20
centered navbar correctly

Gunkev:
In this PR  I centered the elements vertically in the navigation bar
# Original camel website navbar

Screenshot from 2020-03-20 23-52-53

# Camel navbar after changes

Screenshot from 2020-03-20 23-52-30

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -35,6 +35,7 @@ body {
   .navbar-end > .navbar-item,
   .navbar-end .navbar-link {
     color: var(--navbar-font-color);
+    line-height: 55px;
   }
 
   .navbar-end > a.navbar-item:hover,
@@ -45,6 +46,7 @@ body {
 
   .navbar-end .navbar-link::after {
     border-color: var(--navbar-font-color);
+    margin-top: -6px;
   }
 
   .navbar-item.has-dropdown:hover .navbar-link {
@@ -253,7 +255,7 @@ body {
   .navbar-item,
   .navbar-link {
     align-items: flex-end;
-    display: flex;
+    display: block;
     font-weight: bold;
   }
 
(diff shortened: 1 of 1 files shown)


djencks:
Personally I prefer the unmodified proportions.  They are closer to the golden ratio.  In addition, to me it seems that the word "Camel" on the left, being the largest and heaviest visual element, defines where the line is; the unmodified version continues on that line, more or less, while the modified version floats up and doesn't relate to the beginning of the line.

zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-247/Preview/index.html).

antora-ui-camel/src/css/header.css
```diff
@@ -253,7 +255,7 @@ body {
   .navbar-item,
   .navbar-link {
     align-items: flex-end;
-    display: flex;
+    display: block;
```
zregvart: I'm pretty sure that changing this to block from flex breaks something.

antora-ui-camel/src/css/header.css
```diff
@@ -45,6 +46,7 @@ body {
 
   .navbar-end .navbar-link::after {
     border-color: var(--navbar-font-color);
+    margin-top: -6px;
```
zregvart: I'd prefer not to use tricks like negative margins.

antora-ui-camel/src/css/header.css
```diff
@@ -35,6 +35,7 @@ body {
   .navbar-end > .navbar-item,
   .navbar-end .navbar-link {
     color: var(--navbar-font-color);
+    line-height: 55px;
```
zregvart: We prefer not to use px but rather rem as the font size can be changed by the user using the browsers zoom in-out feature.

zregvart:
I don't think this is the right approach. I think we should close this.

PR #237, opened 2020-03-18
Frontpage/projects

shreyaag770:
Rearranged the alignment under Project Section on Frontpage.

q1

PR #250, opened 2020-03-21
rearranged camel-product section

Gunkev:
In this PR  I raarranger and reduce font-size the camel project section below the packed with functionalities to render the latter more readable

# Initial look
Screenshot from 2020-03-21 01-46-37

# Changes made

Screenshot from 2020-03-21 01-45-46

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -156,7 +156,9 @@ section.frontpage.projects .project {
 
 section.frontpage.projects .project h1 {
   white-space: nowrap;
-  padding: 1rem 0;
+  padding: 0;
+  font-size: 22px;
+  text-align: left;
   flex: 1;
 }
 


djencks:
Personally, I find the original easier to understand.  With the rearrangement, I want to know what Camel K being to the right compared to Camel Quarkus underneath mean, compared with Camel Kafka Connector off in the corner.

Gunkev:
In  a website the design most be consistent and should be easy to read or navigate and also be well align to attract the eye . The camel k and other camel is a way group each One in It's container. It ease navigation and user experience rather than going right at the corner to click on read documentation the user will Just move beneath Where his cursor is found and perform his action

Gunkev:
Align all element to left and same top base line, reduce text size and

Screenshot from 2020-03-21 19-21-10

# Changes

Screenshot from 2020-03-21 19-20-35

djencks:
This is less extreme than the previous suggestion, but I personally still find this less effective than the current layout.  To me, the weight of the titles in the current layout, and their centered position both provide visual balance to the text and orange button.  I also like the lower position of the titles in the current layout.
I wonder what positioning the titles at the golden ratio (5:8) relative to the paragraph would look like.  I also wonder what right aligning the titles would look like.
I'd appreciate a sensation based description of the problems you are trying to solve.  There are an unlimited number of possible changes, but a lot of them are not improvements.

shreyaag770:
@djencks I have already sent PR  #237 for right alignment of titles, please have a look at it.

zregvart:
Do we need this or can this be a part of #237?

djencks:
I prefer the overall effect of right-alignment as in #237.

oscerd:
I prefer #237, I don't know why I cannot remove the approval from this one.

zregvart:
Let's go with #237/#264 on this.

PR #254, opened 2020-03-21
CAMEL-14695: Improve Mobile Main-Menu

aashnajena:
- Added drop-downs for menu items having sub-menus, so that all main-menu items are visible in first glance. 
- Made menu full length and adjusted width
- Fixed padding issues

new-mobile-menu

### DIFF
--- a/antora-ui-camel/src/js/04-mobile-navbar.js
+++ b/antora-ui-camel/src/js/04-mobile-navbar.js
@@ -11,13 +11,18 @@ document.addEventListener('DOMContentLoaded', function () {
   })
 
   var menuDropDowns = Array.prototype.slice.call(document.querySelectorAll('.has-dropdown'), 0)
+  var maxMobileWidth = 1023
   menuDropDowns.forEach(function (el) {
     el.addEventListener('click', function (e) {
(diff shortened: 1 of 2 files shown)


antora-ui-camel/src/js/04-mobile-navbar.js
```diff
@@ -9,4 +9,21 @@ document.addEventListener('DOMContentLoaded', function () {
...
+        if (dropDownMenu.style.display === 'block') {
+          dropDownMenu.style.display = 'none'
+        } else {
+          dropDownMenu.style.display = 'block'
+        }
+      } else {
+        dropDownMenu.style.display = 'none'
+      }
+    })
+  })
```
zregvart: This I'm not to keen on. I'd prefer if we just toggeled the class, the Anotra UI already does. Then we would not have to put the width here but we can have the CSS rules depending on the resolution determine the style. i.e. display can be toggled by CSS rules only.

aashnajena: I agree, I hope the changes are fine now. I have toggled the class and added the necessary css to change the display and invert the arrow.

zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-254/Preview/index.html).

PR #259, opened 2020-03-22
resolves camel-14768: fix .htaccess targets

djencks:
I wrote something to generate these changes, and they look good to me, but I'm not sure how to test them.


oscerd:
We need to deploy this change.

PR #260, opened 2020-03-23
Add a search lookup for user-manual page similar to component page

AemieJ:
* It is observed that on the components page, on the side nav panel, there is a lookup search so that the user can easily search within the components list or EIP, based on the query given by the user, the link is easily found instead of dragging through the entire side panel. 

* Thus, for improvement I added the search lookup for the user-manual page as well. Hence, when the user clicks on any sub-topic and based on it searches their query, the particular link is provided making it easy for the user to move through the documents within user-manual.

### DIFF
--- a/antora-ui-camel/src/partials/nav-menu.hbs
+++ b/antora-ui-camel/src/partials/nav-menu.hbs
@@ -3,6 +3,9 @@
   {{#if (eq page.component.name 'components')}}
   <input class="search" placeholder="Quick lookup">
   {{/if}}
+  {{#if (eq page.component.name 'manual')}}
+  <input class="search" placeholder="Quick lookup">
+  {{/if}}
   <nav class="nav-menu">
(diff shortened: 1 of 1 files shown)


AemieJ:
> I think these are two different use cases, the manual is in the hierarchy and the components are listed in the component reference. For searching through the website we have the search functionality.
> I don't think we need this.

Alright, I had a query, why is there a separate search lookup within the components section?

djencks:
The components nav forms a very long flat list of similar but independent things.  The word describing the component functionality might not appear at the start of the item name.  The component search helps find the component you are interested in.  The whole-site search is apt to show things that aren’t components. To me, the component search is not essential but quite handy.

As Zoran already said, the User Manual component is intended to be more narrative, you are very unlikely to want to search for a word in a page title; you’d either look up your topic hierarchically using the nav menus or search for some keyword in the text using the whole-site search.

David Jencks

PR #262, opened 2020-03-23
FIX: nav-menu items overflow issue

AghaSaad04:

## Issue:
Frequently Asked Questions in the nav menu has text that overflows its container and user needs to scroll into x-axis to read the whole text:

image

## Solution:
The text should be divided into multiple lines instead of overflowing.

image


zregvart:
Thanks!

PR #241, opened 2020-03-19
CAMEL-14745: Navbar doesn't fit the screen resolved for few pages

AemieJ:
* The purpose of the PR is to create all the pages within the website responsive irrespective of the screen width.
* It was observed that certain webpages under smaller screen width, the entire webpage doesn't fit the screen and the user needs to scroll through the screen to access the navbar. It is observed in several webpages. 

Currently, I have resolved this issue for
*  **Downloads page**
* **Projects section all page**
* **Blog page and all the articles within it**
*  **Front-page**
* **Community page**
* **User manual pages**
* **EIP**
* **Migration and upgrade guide**

### ISSUE OBSERVED IN DOWNLOADS PAGE AS THE TABLE WAS CREATING THE ISSUE.

Download page issue

### SOLUTION TO RESOLVE THE ISSUE BY WRAPPING THE TABLE UNDER THE TABLE-WRAPPER

Issue resolved using table-wrapper for Download page

PR #267, opened 2020-03-24
[CAMEL-13800] made download page responsive

shreyaag770:
I updated the download page for better mobile usability, hence making it responsive.

Before:

d1

After:

d2


AemieJ:
I am already working on this issue PR#241 and it makes download page along with other pages responsive as well.

zregvart:
We'll go with changes in #241, they're smaller and achieve the same goal AFAIK.

PR #276, opened 2020-03-27
Color consistency in headings of blog page

rimshach:
I have removed the link of the headings  on blog page to make color consistent
image

Furthermore, when we opened individual blog page , the page's heading was a link however it  redirected us to the same page.So I have removed the unnecessary link from those headings as well.
image
The heading shown in the above image redirected to the same page.
Now it looks like this :
image


zregvart:
Preview is [here](https://builds.apache.org/job/Camel.website/job/PR-276/Preview/blog/)

zregvart:
Thanks!

PR #207, opened 2020-03-12
[CAMEL-14700] Improve colour contrast

aashnajena:
- drop-down submenus turn grey on hovering
- put an outline around the search bar to make it stand out
- added a grey shaded gradient (shadow) below nav-bar
- tried changing the background of the highlighted text to darker colours, but it was looking very odd, so I instead changed the colour of the text to camel orange - please comment on this. I have attached screenshots for both cases - background colour and orange text
- menu-items turn dark blue on hovering and while drop downs are open

color_contast_latest

How the highlights look :

camel_highlight_1

And how the bold orange text looks :
highlight_orange
highlight_orange_2

PR #282, opened 2020-03-28
Fix the broken header menu

aashnajena:
As observed in PR #207 the submenus don't start from below the line. I think there was a PR about shifting the menu text to align with the other elements which caused this change. 
- Instead of the margin, I have added padding. Padding left is 0 so that the drop-downs are aligned with the starting of the text
- Added some padding to the down-arrow button, which seemed too close to the text.   

I think the problem in #207 should also be solved with this.

menu-fixed

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -128,6 +128,11 @@ body {
   position: relative;
 }
 
+.navbar-item {
+  flex: none;
+  padding: 1rem 2rem 1rem 0;
+}
+
 .navbar-item .icon {
   width: 1.1rem;
   height: 1.1rem;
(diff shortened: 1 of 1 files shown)


aashnajena:
The drop-down portion starts from below the grey line, so I can make the line beneath the main-menu item white, but it only spans the main-menu-text, not the entire width of the dropdown box.

aashnajena:
Rebased!

zregvart:
This is how it looks currently:

Screenshot_2020-04-01 Outreachy update - Apache Camel

For reference I think what we want is what we [had a month ago](https://web.archive.org/web/20200213140144/http://camel.apache.org/).

Screenshot_2020-04-01 Home - Apache Camel

aashnajena:
I changed it!

aashnajena:
Resolved in PR #306.

PR #277, opened 2020-03-27
make the split in Apache & opensource look neat on front-page

AemieJ:
* Within the **Apache & Opensource** section on the front page, the paragraph written doesn't have any margin surrounding it making the design look unstructured for the mobile view. 

* Thus, I added a margin surrounding the split session for the front page to look structured and neat for the mobile view.

### BEFORE 
before-design

### AFTER
responsive

PR #264, opened 2020-03-24
Changed alignment under Project section on frontpage

shreyaag770:
changes asked in PR  #237 is done in this PR.

PR #283, opened 2020-03-28
Fixed front page issues

aashnajena:
Issue 1: Camel Kafka connecter section is not aligned with the other project sections. I understand that we would prefer to have the titles appearing in single lines to maintain even-ness but this misalignment seems to distort the whole design. 

camel-box-alignment

Issue 2: This section's text does not have any padding on the sides for mobile screens, and according to me, similar to other sections, it would look better if center aligned.

camel_frontpage2

Issue 3: This section has a very odd alignment for tablet view. Since other sections like projects are becoming single column and center aligned, I think this section should follow the same design.

camel_frontpage1

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -160,6 +160,7 @@ section.frontpage.projects .project {
 }
 
 section.frontpage.projects .project h1 {
+  text-align: right;
   padding: 0 1rem;
   flex: 1;
 }
@@ -203,7 +204,7 @@ section.frontpage a:hover {
 
 section.frontpage h1 {
   width: 100%;
-  text-align: right;
+  text-align: center;
 }
 
 section.frontpage h2 {
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -246,7 +246,7 @@ section.frontpage h2 {
 @media screen and (max-width: 626px) {
   .split {
     flex: 100%;
-    margin: 0 2rem;
+    margin: 0 1rem;
   }
 
   section.frontpage.columns,
@@ -270,7 +270,7 @@ section.frontpage h2 {
   }
 
   .split {
-    padding: 0 2rem;
+    margin: 0 1rem;
     flex: 100%;
   }
 
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -160,11 +160,8 @@ section.frontpage.projects .project {
 }
 
 section.frontpage.projects .project h1 {
-  white-space: nowrap;
-  padding: 1rem;
-  flex: 1.2;
-  text-align: right;
-  line-height: 0;
+  padding: 0 1rem;
+  flex: 1;
 }
 
 section.frontpage.projects .project p {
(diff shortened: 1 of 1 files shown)


AemieJ:
I already made the fix to issue #2 of yours in PR #277. Please check next time too before performing the same thing, thankyou very much.

aashnajena:
Sorry for repeating some changes! Although, I don't think we need a margin all around the section. The image and the heading are related, so I'm not very convinced if we need extra margin there. Also, I was referring to the other sections on the front page - the sections about the camel projects all become single column from double columns and become center-aligned, so I'm trying to follow that design. I honestly felt it looks a little odd for the heading to be left aligned beneath the center-aligned image.

shreyaag770:
I have already sent a PR #264 for issue1

zregvart:
I don't think this is an issue any more, can you double check?

AemieJ:
> I don't think this is an issue any more, can you double check?

As long as I have observed, the first issue has been resolved. For the second issue, there is this PR and PR #277  resolving it.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -257,6 +257,16 @@ section.frontpage h1 {
     height: 10rem;
   }
 
+  .split {
+    padding: 0 2rem;
+    flex: 100%;
+    text-align: center;
```
zregvart, line 263: Not sure about aligning the text. If you'd like to center the 20 years of ASF image perhaps it's better to change the `.split img` to `display: block; margin: auto`.

Currently
Screenshot_2020-04-01 Home - Apache Camel(1)

Non centered text and centered image
Screenshot_2020-04-01 Home - Apache Camel(2)

aashnajena: Okay, if left-aligned is preferred I'll change that!

aashnajena:
Also, @zregvart I'm looking at the current website and issue 1 does not seem to be resolved for me on all browsers.
 
align

AemieJ:
> Also, @zregvart I'm looking at the current website and issue 1 does not seem to be resolved for me on all browsers.
> 
> align

If the browser size is large enough, it is going to appear like that. Only for the smaller screen width, it will easily customize and none of them are perfectly aligned with each other.

aashnajena:
Thanks for clarifying @AemieJ! The problem I was trying to solve still persists, that's all I'm saying. It should be solved and the alignment should be proper for all screen sizes. Also, @shreyaag770's PR seems to be closed. I'll make the requested changes for the Apache and Opensource section soon.

aashnajena:
rebased to add the right alignment of project titles, while keeping the alignment intact. Also added requested changes.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -248,6 +246,7 @@ section.frontpage h2 {
 @media screen and (max-width: 626px) {
   .split {
     flex: 100%;
+    margin: 0 2rem;
```
zregvart, line 249:
```suggestion
    margin: 0 1rem;
```

antora-ui-camel/src/css/frontpage.css
```diff
@@ -270,6 +269,17 @@ section.frontpage h2 {
     height: 10rem;
   }
 
+  .split {
+    padding: 0 2rem;
```
zregvart, line 273:
```suggestion
    margin: 0 1rem;
```

aashnajena: yes, I think the margin looks a bit much for the mobile screen. Made the suggested changes.

zregvart:
I've rebased this to get it to merge. Thanks!

PR #284, opened 2020-03-28
Alignment of navbar-link:after arrows

psingh587:
This is how the arrows look in the mobile screen resolutions. My PR serves to fix this.

WhatsApp Image 2020-03-29 at 06 58 36

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -225,7 +225,7 @@ body {
     transform: rotate(-45deg);
     width: 0.5em;
     right: 0;
-    top: 40%;
+    top: 4.5px;
   }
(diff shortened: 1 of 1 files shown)


antora-ui-camel/src/css/header.css
```diff
@@ -225,7 +225,7 @@ body {
     transform: rotate(-45deg);
     width: 0.5em;
     right: 0;
-    top: 40%;
+    top: 4.5px;
```
zregvart: We can't use pixels, we need to use `rem`. There is no 0.5 pixel size on the screen.

psingh587:
Seems like the issue has been fixed. I am closing the PR.

PR #222, opened 2020-03-16
Fixed blog page categories issue

jyotiattri:
Implemented the requested changes from [#213 ](https://github.com/apache/camel-website/pull/213)

PR #285, opened 2020-03-29
Include the aside block for post but not for blog list

AemieJ:
* I observed that the previous and next arrow direction within the articles were not being displayed in the mobile view. However, when you read a blog article I feel it's very inconvenient to keep moving back to the blog page search through the article and then reading it. Thus, I felt the use of the **previous** and **next** arrows were essential. 

* Hence, I made changes such that the blog list is only hidden for the mobile view however the tag, author and directions of the page is included for each blog post.

### DIFF
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -106,6 +106,10 @@ article.blog p {
   margin-bottom: 1rem;
 }
 
+.blog .post .post-content .direct {
+  display: none;
+}
+
 .blog .post .post-content tr:nth-child(even) {
   background: var(--color-smoke-50);
 }
@@ -159,5 +163,11 @@ article.blog p {
 @media screen and (max-width: 1023px) {
   .blog aside {
     display: none;
+  }
+
+  .blog .post .post-content .direct {
+    display: block;
+    text-align: center;
+    margin-top: 2rem;
   }
 }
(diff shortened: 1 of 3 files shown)


AemieJ:
> I think this issue was caused by #222, perhaps better approach would be to hide the `aside` in `.blog.list` only.

The **aside** block doesn't get displayed in the mobile view. So I thought it was better to provide the user with **previous** and **next** pages for ease. What are you suggesting the better approach is?

zregvart:
> The **aside** block doesn't get displayed in the mobile view. So I thought it was better to provide the user with **previous** and **next** pages for ease. What are you suggesting the better approach is?

Previous and next links of single blog post are part of the `aside` which got hidden in #222 as we wanted to hide the categories menu in the blog post list. We could show `aside` in single blog post but still hide it in blog list.

AemieJ:
> > The **aside** block doesn't get displayed in the mobile view. So I thought it was better to provide the user with **previous** and **next** pages for ease. What are you suggesting the better approach is?
> 
> Previous and next links of single blog post are part of the `aside` which got hidden in #222 as we wanted to hide the categories menu in the blog post list. We could show `aside` in single blog post but still hide it in blog list.

It's a better option to include the summary within the single blog post. However, in my opinion I feel that the user will prefer the next and previous arrow links at the end of the document. I will try to find a better approach and see.

zregvart:
Thanks!

PR #287, opened 2020-03-29
Scrollbar hiding issue fixes

psingh587:
Earlier with 100% Quick Look up search box width, the following problem occured. The scroll bar used to get hidden beneath it. :

ezgif com-video-to-gif (1)

After changes :- 

ezgif com-video-to-gif


psingh587:
> Better way would be to make the `nav-menu` scrollable instead.

Done as asked.

zregvart:
Thanks!

PR #290, opened 2020-03-30
fix table alignment issue within Team page

AemieJ:
## ISSUE 

* The table alignment was improper within the team page. Hence, in the **team.md**, I wrote the table in a structure and neat manner and it solves the alignment issue for the table within the **Teams** page. 

### BEFORE 
team-before

### AFTER
team-after


AemieJ:
> This should be splitted in two PRs

This PR is now concerned with the proper alignment of the table itself only.

PR #294, opened 2020-03-31
CAMEL-14581: improve accessibility ratio through color contrast under blog section

AemieJ:
* After running audit, it was observed that the accessibility factor under the blog section was decreasing due to various factors, one of them including the **color contrast**. 

* For the case of  **color contrast**, I have made all required changes to get the right **COLOR CONTRAST RATIO** while keeping in mind the aesthetic and presentability of the pages under the blog section.


zregvart:
Thanks!

PR #295, opened 2020-03-31
Custom Pagination within Blog Section

AemieJ:
I replaced the template of the pagination of Hugo with custom pagination. In this custom pagination, it has the following properties: 

1. Show **<<<<** shortcut if not 1st page

page-start

2. Show **>>>>** shortcut if not the last page.

page-end

3. Show **<<** if there is a previous page
4. Show **>>** if there is the next page
5. Show current page number with the previous 2 pages and next 2 pages. Skip other pages with placeholder ...

page-middle

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -368,7 +368,7 @@
 .doc ol,
 .doc ul {
   margin: 0;
-  padding: 0 0 0 2rem;
+  padding: 0 0 0 1rem;
 }
 
 .doc ol.arabic {
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -368,7 +368,7 @@
 .doc ol,
 .doc ul {
   margin: 0;
-  padding: 0 0 0 1rem;
+  padding: 0 0 0 2rem;
 }
 
 .doc ol.arabic {
--- a/layouts/blog/list.html
+++ b/layouts/blog/list.html
@@ -18,37 +18,43 @@
             
             {{ $page := .Paginator }}
             {{ if gt $page.TotalPages 1 }}
-            {{ $.Scratch.Set "dot_rendered" false }}
-            <nav aria-label="page navigation">
+            {{ $shouldEllipse := false }}
+            <nav aria-label="page">
                 <ul class="pagination">
-                    <!-- Don't show on 1st and 2nd page -->
-                    {{ if and (ne $page.PageNumber 1) }}
-                    <li class="page-item"><a href="{{ $page.First.URL }}" rel="first" class="page-link">««</a></li>
+                    {{ if ne $page.PageNumber 1 }}
+                    <li class="page-item">
+                        <a href="{{ $page.First.URL }}" class="page-link" aria-label="First"><span aria-hidden="true">&laquo;&laquo;</span></a>
+                    </li>
                     {{ end }}
 
                     {{ if $page.HasPrev  }}
-                    <li class="page-item"><a href="{{ $page.Prev.URL }}" rel="prev" class="page-link">«</a></li>
+                    <li class="page-item">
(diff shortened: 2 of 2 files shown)


antora-ui-camel/src/css/doc.css
```diff
@@ -368,7 +368,7 @@
 .doc ol,
 .doc ul {
   margin: 0;
-  padding: 0 0 0 2rem;
+  padding: 0 0 0 1rem;
```
zregvart, line 371: This change doesn't seem to be related to pagination. If it is the rule needs to be scoped to the pagination and placed in blog.css. Otherwise it affects other `ol` and `ul` elements.

layouts/blog/list.html
```diff
@@ -15,9 +15,46 @@ <h3>Categories</h3>
             {{ range $pages }}
                 {{ .Render "summary" }}
             {{ end }}
-            {{ template "_internal/pagination.html" . }}
+            
+            {{ $page := .Paginator }}
+            {{ if gt $page.TotalPages 1 }}
+            {{ $.Scratch.Set "dot_rendered" false }}
+            <nav aria-label="page navigation">
```
zregvart, line 22: Not sure we can use multiple strings for `aria-label`, [specification says](https://www.w3.org/TR/wai-aria/#aria-label) value in singular

layouts/blog/list.html
```diff
@@ -15,9 +15,46 @@ <h3>Categories</h3>
...
+                <ul class="pagination">
+                    <!-- Don't show on 1st and 2nd page -->
```
zregvart, line 24: Use [Hugo comments](https://gohugo.io/templates/introduction/#comments), no use of this in the rendered markup

layouts/blog/list.html
```diff
@@ -15,9 +15,46 @@ <h3>Categories</h3>
...
+                    {{ if and (ne $page.PageNumber 1) }}
```
zregvart, line 25: Perhaps

```suggestion
                    {{ if not .$page.First }}
```

layouts/blog/list.html
```diff
@@ -15,9 +15,46 @@ <h3>Categories</h3>
...
+                    {{ end }}
+                </ul>
+            </nav>
+            {{ end }}
         </div>
     </main>
 </div>
 
-{{ partial "footer.html" . }}
+{{ partial "footer.html" . }}
```
zregvart, line 60: No newline at EOF

layouts/blog/list.html
```diff
@@ -15,9 +15,46 @@ <h3>Categories</h3>
...
+                    <li class="page-item"><a href="{{ $page.Next.URL }}" rel="next" class="page-link">»</a></li>
+                    {{ end }}
+
+                    <!-- Don't show on last and 2nd last page -->
+                    {{ if and (ne $page.PageNumber $page.TotalPages) }}
+                    <li class="page-item"><a href="{{ $page.Last.URL }}" rel="last" class="page-link">»»</a></li>
+                    {{ end }}
+                </ul>
+            </nav>
+            {{ end }}
```
zregvart, line 55: 1. We're missing `aria-label` in `a` tags. Examine the [default template](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/pagination.html).
2. Use `.Last` as `.First` above
3. Avoid using scratch if possible, use explicit conditions or local variables instead (consult the internal Hugo template)

AemieJ: Alright, I will make use of this and do the required changes.

layouts/blog/list.html
AemieJ: The command you suggested isn't performing. **not $page.First** doesn't provide any output.

layouts/blog/list.html
AemieJ: I have done except the 2nd one I had to use the **ne** condition as **not** wasn't given the expected results.

zregvart:
Thanks!

PR #297, opened 2020-03-31
chnaged alignment of headings under packed with fuctionality section

shreyaag770:
I changed the alignment of headings under "packed with functionality" section for better UI

Preview :
(Destop view):
p1

(Mobile view):
p2


zregvart:
Thanks!

PR #272, opened 2020-03-25
 Pull contributor data from JSON APIs

CynthiaOmovoiye:
CAMEL 13894- Pull contributor data from JSON APIs

I have been able to pull all required data from the JSON files into the team page for Committers. I have also been able to set the GitHub API that extracts the collaborators data from the repository and rendered it to display on the team page. However, it will require a GitHub access token from the Apache GitHub account for the data to be displayed but I was able to generate one on my forked repository of the Apache camel website for testing. Thus, if an access token  generated by an admin of the Apache camel website repository is inputted in the field I have made available for it in(" /layouts/shortcodes/contributors.html " **{{$token :=" "}}**), we would be able get the data of the collaborators  displayed on the web page.
.

PR #298, opened 2020-03-31
Pulling data from json apis

CynthiaOmovoiye:
This PR addresses and fixes the corrections made from PR #272 .


CynthiaOmovoiye:
The Contributors table in Contents\Community\ team.md now pulls its content from Github API for contributors; displaying the avatars of the contributors alongside their names . However, the names of some contributors where absent on the repository have been replaced with their "login ID", which if deemed unnecessary can be changed. Also, the names of anonymous contributors have also been omitted to prevent repetition of names and empty fields, if these information are found necessary, we can re-incorporate them. 

contributors4

config.toml
```diff
@@ -12,6 +12,7 @@ timeout = 300000
     socialProfiles = ["https://twitter.com/ApacheCamel"]
     organizationLogo = "https://camel.apache.org/_/img/logo-d.svg"
     organizationDescription = "Apache Camel ™ is a versatile open-source integration framework based on known Enterprise Integration Patterns. Camel empowers you to define routing and mediation rules in a variety of domain-specific languages, including a Java-based Fluent API, Spring or Blueprint XML Configuration files, and a Scala DSL."
+    token = "3656a8b91e08cbd62a3d92ba09360a8d0f233216"
```
zregvart, line 15: **Please remove this and revoke this token.** Not sure how you scoped this token but it could be used instead of your password in your name!

content/community/team.md
```diff
@@ -8,169 +8,18 @@ This page lists who we are. By all means add yourself to the list - lets sort it
...
+
+
+
 
+
+## Contributors
 Adding your name to the list below.
 
-If you have been contributing to the Apache Camel project, and you want your name added to the list below. Then you can get in touch with the Camel team from the [Mailing Lists](../mailing-list/) and ask to be added.
+If you have been contributing to the Apache Camel project, and you want your name added to the list below. Then you can get in touch with the Camel team from the <a href="../mailing-list/">Mailing Lists</a> and ask to be added.
```
zregvart, line 21: Leave the markup as it is. No need to add HTML markup for links here.

data/teamdata.json
```diff
@@ -0,0 +1,61 @@
...
+    "acosentino": "Red Hat",
+    "astefanutti": " Red Hat",
+    "bvahdat": "Cyberlogic Consulting GmbH",
+    "boday": "Initek Consulting",
+    "bibryam": "Red Hat",
+    "bsnyder": " ",
+    "cmoulliard": "Red Hat",
+    "cmueller": "Amazon Web Services",
+    "ceposta": "Red Hat",
+    "cschne\"id\"er": "",
```
zregvart, line 13: Looks mangled

layouts/shortcodes/contributors.html
```diff
@@ -0,0 +1,40 @@
...
+    <tbody>
+        {{ range  $page_count := (seq 10) }}
+            {{ $page := string $page_count }}
+                {{ $git_contributors := getJSON "https://api.github.com/repos/apache/camel/contributors?per_page=100&anon=1&page=" $page "&access_token=" $.Site.Params.token }}
+                {{ range  $git_contributors }}
+                    {{ if .url }}
+                        {{ $user_url := getJSON .url }}
+                        {{ if $user_url.name }}
+                            <tr>
+                                <td> <img width="50px" height="50px" src="{{ .avatar_url }}"/></td>
```
zregvart, line 19: Would prefer to use a CSS class instead of `width`/`height`

layouts/shortcodes/contributors.html
```diff
@@ -0,0 +1,40 @@
...
+        </tr>
+    </thead>
+    <tbody>
+        {{ range  $page_count := (seq 10) }}
+            {{ $page := string $page_count }}
+                {{ $git_contributors := getJSON "https://api.github.com/repos/apache/camel/contributors?per_page=100&anon=1&page=" $page "&access_token=" $.Site.Params.token }}
+                {{ range  $git_contributors }}
+                    {{ if .url }}
+                        {{ $user_url := getJSON .url }}
+                        {{ if $user_url.name }}
```
zregvart, line 17: Use [`default`](https://gohugo.io/functions/default/) instead of branching

layouts/shortcodes/contributors.html
```diff
@@ -0,0 +1,40 @@
...
+            <th>Name</th>
+            <th>Organization</th>
+        </tr>
+    </thead>
+    <tbody>
+        {{ range  $page_count := (seq 10) }}
+            {{ $page := string $page_count }}
+                {{ $git_contributors := getJSON "https://api.github.com/repos/apache/camel/contributors?per_page=100&anon=1&page=" $page "&access_token=" $.Site.Params.token }}
+                {{ range  $git_contributors }}
+                    {{ if .url }}
```
zregvart, line 15: Nicer way would be to use `with`

config.toml
CynthiaOmovoiye: fixed

content/community/team.md
CynthiaOmovoiye: fixed

data/teamdata.json
CynthiaOmovoiye: fixed

layouts/shortcodes/contributors.html
CynthiaOmovoiye: fixed

layouts/shortcodes/contributors.html
CynthiaOmovoiye: fixed

layouts/shortcodes/contributors.html
CynthiaOmovoiye: okay fixed

layouts/shortcodes/contributors.html
```diff
@@ -0,0 +1,31 @@
...
+<table class="tableblock stretch">
+    <thead>
+        <tr>
+            <th>Avatar</th>
+            <th>Name</th>
+            <th>Organization</th>
+        </tr>
+    </thead>
+    <tbody>
+        {{ range  $page_count := (seq 10) }}
```
zregvart, line 11: Option 1: comment so we know what this is about
```suggestion
        {{/* TODO we can't access HTTP headers and to fetch all contributors we need to load multiple pages, when Camel has over 1000 contributors the 10 pages we have here will not be enough */}}
        {{ range  $page_count := (seq 10) }}
```

Option 2: do a `range` over much bigger sequence and don't invoke `getJSON` if last invocation returned 100 contributors, something like:
```
{{/* we're looping over 100 numbers in order to fetch paged data from GitHub API, we're invoking GitHub API only if we received all results per page (100) from the last page */}}
{{ $pages := seq 100 }}¬
{{ $last_contributors := 1 }}¬
{{ range $page, $tmp := $pages }}¬
    {{ if eq $last_contributors 100 }}¬
        {{ $git_contributors := getJSON "https://api.github.com/repos/apache/camel/contributors?per_page=100&anon=1&page=" ($page | string) }}¬
        {{ $last_contributors = len $git_contributors }}¬
...
    {{ end }}¬
{{ end }}¬

```

antora-ui-camel/src/css/doc.css
```diff
@@ -644,6 +644,10 @@ kbd,
   width: 100%;
 }
 
+.git-img {
+  width: 50px;
+  height: 50px;
+}
```
zregvart, line 650: Use this:

https://github.com/apache/camel-website/blob/60537f0c4d18dce038d2922a72644cc9342cf6ef/antora-ui-camel/src/css/blog.css#L128-L130

Put this change in `community.css` and use `.community img.avatar` as rule.

layouts/shortcodes/contributors.html
```diff
@@ -0,0 +1,31 @@
...
+            {{ $page := string $page_count }}
+                {{ $git_contributors := getJSON "https://api.github.com/repos/apache/camel/contributors?per_page=100&anon=1&page=" $page "&access_token=" $.Site.Params.token }}
+                {{ range  $git_contributors }}
+                    {{ with .url }}
+                        {{ $user_url := getJSON . }}
+                            <tr>
+                                <td> <img class="git-img" src="{{ $user_url.avatar_url }}"/></td>
```
zregvart, line 18:
```suggestion
                                <td> <img class="avatar" src="{{ $user_url.avatar_url }}"/></td>
```

config.toml
```diff
@@ -12,6 +12,7 @@ timeout = 300000
     socialProfiles = ["https://twitter.com/ApacheCamel"]
     organizationLogo = "https://camel.apache.org/_/img/logo-d.svg"
     organizationDescription = "Apache Camel ™ is a versatile open-source integration framework based on known Enterprise Integration Patterns. Camel empowers you to define routing and mediation rules in a variety of domain-specific languages, including a Java-based Fluent API, Spring or Blueprint XML Configuration files, and a Scala DSL."
+    token = ""
```
zregvart, line 15: Remove this completely, we're using public APIs, until we hit rate limits we don't need this.

layouts/shortcodes/contributors.html
CynthiaOmovoiye: This outputs a blank table[img1] but when `$last_contributors := 100` it works but with an API limit error[error]
img1


error

config.toml
CynthiaOmovoiye: I have removed the token but the page returns an error[error]. I think it has to do with API rate limit. What do you think?

error

zregvart:
This turned out more complex than it's worth, with API rate limits on GitHub we are going to run into unexpected build issues. I'll close this in a few days if no one objects.

PR #301, opened 2020-04-01
Added an Anchor tag to reach the Top of the page

psingh587:
- I had created a PR #234 before this. There were some file glitches which I was unable to eradicate in that PR so I created this new one with all the requested changes. 
      **1. "Back to Top" tag added at the bottom of page**
      **2. For pages under Documentation, the "Back to top" is put in the TOOLBAR **


antora-ui-camel/src/css/frontpage.css
```diff
@@ -207,7 +207,7 @@ section.frontpage h1 {
 }
 
 section.frontpage h2 {
-  text-align:center;
+  text-align: center;
```
psingh587, line 210: This is unrelated to the objective of my PR. But I had to make this change because yarn build was throwing an error.

antora-ui-camel/src/partials/footer-content.hbs
```diff
@@ -1,4 +1,4 @@
-    <footer>
+  <footer>
```
psingh587, line 1: Here, earlier I had put the <div class="top".. . Now I have removed it and for the Documentation pages, as mentioned previously, the Back to Top tag is put in the toolbar within the same lines as Edit The Page and Wide/Narrow.

antora-ui-camel/src/css/frontpage.css
zregvart: Yeah I've noticed that, fixed on master in 3a11366d3d70577c971bc151deab9dd1fbfea51b

antora-ui-camel/src/partials/BackToTop.hbs
```diff
@@ -0,0 +1 @@
+<div class="top"><a href="#top">Back To Top </a></div>
```
zregvart, line 1: Not sure we need a partial for this, it's quite small to begin with. If it is needed then it needs to conform to the file naming conventions other partials have.

psingh587: > I'd prefer to put the back to top link in the same place on the Antora built content and as on the Hugo built content.
@zregvart  Actually earlier, I was going with the same thing as you are suggesting, but I encountered a problem as shown below in the documentation pages. I thought maybe that's the reason Edit Page link is also not set at the bottom in the Documentation pages. 


Screenshot from 2020-04-01 19-15-11

psingh587: > Not sure we need a partial for this, it's quite small to begin with. If it is needed then it needs to conform to the file naming conventions other partials have.


Made the changes

psingh587:
A PR #307 has been made based on this. I am closing this PR

PR #303, opened 2020-04-01
Improvize the frontpage design

AemieJ:
* I had some issues regarding the arrangement with the text within the **navbar** and in the **header frontpage**.
1.  In the **navbar**, the navbar links were really close to the logo brand, hence I provided a margin between the logo and the **navbar-end**.

### BEFORE 
alignment-before

### AFTER 
alignment

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -420,6 +420,7 @@ body {
   font-family: Open Sans, sans-serif;
   caret-color: #ed8225;
   background: no-repeat 0.2rem/1.2rem url(../img/search.svg);
+  width: 8.75rem;
 }
 
 .navbar-search input:focus-within {
@@ -486,8 +487,15 @@ body {
   }
 }
 
+@media screen and (min-width: 1024px) and (max-width: 1230px) {
+  .navbar-end {
+    justify-content: flex-end;
+    margin-left: 1rem;
+  }
+}
+
 /* Styling for mobile use */
-@media (max-width: 610px) {
+@media (max-width: 1175px) {
   #search {
     display: none;
   }
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -352,7 +352,7 @@ body {
 
   .navbar-end {
     justify-content: flex-end;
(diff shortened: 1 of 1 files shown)


AemieJ:
Alright, then we can increase the space and leave the text as it is. I will make the changes related to that.

zregvart:
Just noticed that the whole top level navigation doesn't fit on 1024px width, both in the current version and in this PR. It's probably been like that since we added the search box. Not sure if this PR is helping with that though it looks better on wider screens. Not sure how to proceed, seems like we need to rethink the top level navigation.

AemieJ:
> Just noticed that the whole top level navigation doesn't fit on 1024px width, both in the current version and in this PR. It's probably been like that since we added the search box. Not sure if this PR is helping with that though it looks better on wider screens. Not sure how to proceed, seems like we need to rethink the top level navigation.

I can work on that and see if I can resolve that.

AemieJ:
I have additionally done the following changes: 
1. For width <= 1024px, the search box will not appear.
2. For width between 1024px-1230px, the margin between logo and the navbar menu is 1rem. 
3. For default, the margin between the logo and the navbar menu is 2rem.

Delawen:
So for width < 1024px we will not be able to search? I know less than 1024px is not common on computers but it's still like 9% of total web users:

https://www.w3schools.com/browsers/browsers_display.asp

AemieJ:
> So for width < 1024px we will not be able to search? I know less than 1024px is not common on computers but it's still like 9% of total web users:
> 
> https://www.w3schools.com/browsers/browsers_display.asp
> https://gs.statcounter.com/screen-resolution-stats

I will see through this.

AemieJ:
@zregvart  How about for screen width between 1024px and 1175px wi, the search is present below the social icons and the navbar height is increased?

zregvart:
I think this overall improves the design.

There's also a pull request to put the search box in the hamburger menu, so folk with lower widths can still access the search.

In the future I think we need to rethink the number of items in the top level navigation and media breakpoints -- we seem to be adding more and more of them rather than designing for specific screen widths based on data we have. I think this requires bit more effort so best left for the internship where we can devote the time to change the design as we see fit.

PR #304, opened 2020-04-02
Table new design

jyotiattri:
In this PR, I have changed the design of all table that is present on the website. I have changed the design to make UI same for all the tables.
**Before:**
 table-before

**After:**
table-after

Release table:-
**Before:**
relase-table-before

**After:**
relase-table-after

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -203,7 +203,6 @@
 }
 
 .doc table.tableblock td {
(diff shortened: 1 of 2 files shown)


AemieJ:
This looks pretty neat but under the release table, for the header, I think it would look better and consistent if you removed the border-top for the headers.

jyotiattri:
@AemieJ, Yes I also think. Thanks for the comment. I will do that.

zregvart:
Thanks!

PR #306, opened 2020-04-02
fix: top level navigation dropdown positioning

zregvart:
This fixes the positioning of the top level navigation dropdown making
it look seamless with the menu item. Also simplifies the CSS by removing
duplication.

@aashnajena I think we're going to go with this instead of #282 can you have a look?


aashnajena:
Sure @zregvart, this is a better way of doing it than changing percentages. I'll close the other PR.

PR #309, opened 2020-04-03
Straight lines as separator

raashika03:
@zregvart  As you suggested to incorporate few narrow straight lines as a separator beneath the main topmost navigation and the rest of the webpage. I did so.please tell if I'm heading in the right direction or not. Currently I'm working on the transition animation
Screenshot from 2020-04-03 03-46-54

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -69,7 +69,7 @@
                 </button>
             </div>
         </nav>
-        <svg height="10" width="100%">
+        <svg viewBox="0 0 500 70" preserveAspectRatio="none" style="position: absolute;top: 4rem;left: 0;width: 100%;height: 3rem;">
   <line x1="0" y1="0" x2="100%" y2="0" style="stroke:rgb(205,133,63);stroke-width:5"></line>
   <line x1="0" y1="5" x2="100%" y2="5" style="stroke:rgb(210,105,30);stroke-width:5"></line>
   <line x1="0" y1="10" x2="100%" y2="10" style="stroke:rgb(160,82,45);stroke-width:5"></line>
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -69,4 +69,9 @@
                 </button>
             </div>
         </nav>
+        <svg height="10" width="100%">
+  <line x1="0" y1="0" x2="100%" y2="0" style="stroke:rgb(205,133,63);stroke-width:5"></line>
+  <line x1="0" y1="5" x2="100%" y2="5" style="stroke:rgb(210,105,30);stroke-width:5"></line>
+  <line x1="0" y1="10" x2="100%" y2="10" style="stroke:rgb(160,82,45);stroke-width:5"></line>
+</svg>
     </header>


Delawen:
Could we add some blank space so the background image is not seen above the line?

image

zregvart:
:+1: This looks like something I had in mind. Setting this for a bit ambitious effort, for animation I was thinking that the swoosh would transform into those lines as the user scrolled down. So on frontpage we would not have the lines present until the user scrolls down, and as the user scrolls down the swoosh would transform into those lines.

But:exclamation: Please, it need not be like that, feel free to experiment and do what you think is a good option.

And as we don't have the swoosh on the other pages I would use the lines as a separator (probably not animated) between the top level navigation and the page content. Similar to what you have here.

raashika03:
> Could we add some blank space so the background image is not seen above the line?
> image
> 
> Preview on https://deploy-preview-309--camel.netlify.com/

@Delawen As per your suggestion,I removed those spaces
Screenshot from 2020-04-04 06-19-18

raashika03:
>  This looks like something I had in mind. Setting this for a bit ambitious effort, for animation I was thinking that the swoosh would transform into those lines as the user scrolled down. So on frontpage we would not have the lines present until the user scrolls down, and as the user scrolls down the swoosh would transform into those lines.
> 
> But Please, it need not be like that, feel free to experiment and do what you think is a good option.
> 
> And as we don't have the swoosh on the other pages I would use the lines as a separator (probably not animated) between the top level navigation and the page content. Similar to what you have here.
I applied this separator to the whole website,what I noticed is that total three types of formats have been used for the whole website
1) frontpage in which swoosh design is there, no text is written close to the navbar in the body
2)contributing under community heading in which if I'll use the separator  then visibilty issues of the text will be there which are pretty close to the navbar,
3)license under about section which has the navbar of small height, so the separator used in those cases does not seems to appeal, I tried different effects but they didn't seem to provide nice visual effects 
so,I left 2 and  3 as they are
Now for the transition,transforming those curvy lines to the straight lines while scrolling down and since the path of each shade in that swoosh is curve which is not getting covered at the same time by the navbar which will result in the formation of the separator in pieces which won't look good because of the colour contrast  as per my suggestion

layouts/partials/header.html
```diff
@@ -69,4 +69,9 @@
                 </button>
             </div>
         </nav>
+        <svg viewBox="0 0 500 70" preserveAspectRatio="none" style="position: absolute;top: 4rem;left: 0;width: 100%;height: 3rem;">
```
zregvart, line 72: The content of the `style` attribute needs to go to one of the css files, I think `static.css`, otherwise it will not load as we explicitly forbid inline style, see:

https://github.com/apache/camel-website/blob/18028d3ac6756094d67bb383164724d9dc367b9e/static/.htaccess#L1329

You can test this by running the web server in Docker, see https://github.com/apache/camel-website/tree/master/support#running-apache-http

layouts/partials/header.html
```diff
@@ -69,4 +69,9 @@
...
+  <line x1="0" y1="0" x2="100%" y2="0" style="stroke:rgb(205,133,63);stroke-width:5"></line>
+  <line x1="0" y1="5" x2="100%" y2="5" style="stroke:rgb(210,105,30);stroke-width:5"></line>
+  <line x1="0" y1="10" x2="100%" y2="10" style="stroke:rgb(160,82,45);stroke-width:5"></line>
```
zregvart, line 75: Same goes for these `style` attributes, they need to be in a `.css` file

zregvart:
So note on the where this fits, i.e. in what CSS file, if we're going to add this to Hugo built content, i.e. content without the toolbar with breadcrumbs, wide/narrow & edit this page at the top, then it needs to go to `static.css`, if we're going to add this to all pages, including the pages built by Antora (i.e. those with that toolbar) then I think it's best to put it in `header.css`.

PR #310, opened 2020-04-03
Make modifications to the drop-down menu and provide a new design alternative

AemieJ:
### Drop-down menu design alternative
Based on the design alternatives for drop-down I observed on various websites. I made corresponding changes to provide a neat and interactive drop-down menu design.

[Click here for seeing the change](https://github.com/apache/camel-website/files/4425942/drop-down.zip)


AemieJ:
I did the changes and now `yarn build` happens and can be previewed.

zregvart:
I like that you're experimenting. I think it would be good not to loose the indicator that some menus have drop-downs, do you think that would work, something like:

Screenshot_2020-04-03 Apache Camel user manual - Apache Camel(1)

I got this by tweaking the `.navbar-link::after`'s `right: 50%` and `top: 95%`

Not sure about the jumping of text on hover. I'd minimize motion if possible. I also think some of the CSS can be simplified, but let's see about the design and if some other folk have feedback.

AemieJ:
> I like that you're experimenting. I think it would be good not to loose the indicator that some menus have drop-downs, do you think that would work, something like:
> 
> Screenshot_2020-04-03 Apache Camel user manual - Apache Camel(1)
> 
> I got this by tweaking the `.navbar-link::after`'s `right: 50%` and `top: 95%`
> 
> Not sure about the jumping of text on hover. I'd minimize motion if possible. I also think some of the CSS can be simplified, but let's see about the design and if some other folk have feedback.

I will think of a manner to minimize the motions for the hover text. For adding the arrow link for the hoverable menus, I think it being beside the text is better than being below.

zregvart:
Just for some context, I was thinking of adding the little arrow below the text as this would help us in shrinking the amount of space we need horizontally, so we can fit all the elements on smaller screens.

AemieJ:
> Just for some context, I was thinking of adding the little arrow below the text as this would help us in shrinking the amount of space we need horizontally, so we can fit all the elements on smaller screens.

I was planning to create an interactive 3 arrow dropdown hover effect. I am working on it, will see if it fits properly.

AemieJ:
@zregvart for the custom drop down, I added the arrows single design itself and I removed any motion observed on selection.

I tried the designs for on hover 2 or 3 arrows however it doesn't go with the navbar design.

zregvart:
The blur on hover is a bit too much for my taste. I liked that the grey line between the menu item and the pull down menu is not visible on the current design, so the pull down menu looks like it expanded from the menu item.

Screenshot_2020-04-15 Blog - Apache Camel

AemieJ:
> The blur on hover is a bit too much for my taste. I liked that the grey line between the menu item and the pull down menu is not visible on the current design, so the pull down menu looks like it expanded from the menu item.
> 
> Screenshot_2020-04-15 Blog - Apache Camel

I will work on removing the grey line visibility so that it looks expanded from the menu item

zregvart:
@AemieJ are you working on this, this needs to be rebased now.

AemieJ:
I am not working on this currently. I think it's best to close this PR for now.

PR #311, opened 2020-04-03
chore: upgrade Antora to 2.3 beta

zregvart:
Given that we can decrease the build times with newer version of
Antora[1] it makes sense to try it out.

This also cleans up the unneeded dependencies and moves the dependencies
to `devDependencies` section.

[1] https://gitlab.com/antora/antora/-/issues/451

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,25 +15,20 @@
     "checks": "run-s check:*"
   },
   "devDependencies": {
-    "@antora/cli": "~2.1.0",
-    "@antora/site-generator-default": "~2.1.0",
+    "@antora/cli": "2.3.0-beta.2",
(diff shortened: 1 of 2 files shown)


zregvart:
I'm seeing ~20% improvement in build times, @djencks do you know of any issues with upgrading Antora to 2.3 beta by any chance?

djencks:
I know of no issues.  It's possible that some new features might still change a little bit, but since they are new presumably we aren't using them :-)

zregvart:
I've upgraded to 2.3 beta 2. I haven't noticed any side effects, apart from performance improvements. I'm going to merge this if we see any issues it's easy to revert.

PR #312, opened 2020-04-03
make table in security page responsive

AemieJ:
* The table within the security table wasn't responsive in the mobile design. 
* The table is now contained within a **table-wrapper** which makes it responsive in the mobile view with the changes in this PR.


zregvart:
Thanks!

PR #314, opened 2020-04-04
Make modification in README regarding yarn build

AemieJ:
* I observed that in a few cases when I run `yarn build` within the project directory, it throws the error **Javascript heap out of memory**. Hence, I found a solution and I think it will be better to put it in the README documentation for a better understanding for the users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ In case `yarn build` throws the error: **JavaScript heap out of memory**, the is
 
 ```shell
 $ export NODE_OPTIONS="--max_old_space_size=4096"
+```
 This should fetch doc sources for [Camel](https://github.com/apache/camel) and [Camel K](https://github.com/apache/camel-k)
 and generate the website with Hugo. You should see the generated website in the `public` directory.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,6 +137,12 @@ To build the website go to the project root directory and run:
     $ yarn install # needed only once, or if dependencies change
     $ yarn build   # to perform the build
 
+In case `yarn build` throws the error: **JavaScript heap out of memory**, make the following change within `package.json` 
+and the issue will be resolved: 
+```
+"build": "run-s --max_old_space_size=4096 build:*"
+```
+
```
zregvart, line 145: If we need to configure old space for all cases, i.e. the build is broken without this then I would change the `package.json` otherwise if it's an specific issue some contributors might experience but it's not widely and issue for everyone, I'd suggest to set the `NODE_OPTIONS` environment variable:

```suggestion
In case `yarn build` throws the error: **JavaScript heap out of memory**, the issue can be resolved by increasing the memory used by node.js by setting `NODE_OPTIONS` environment variable to include `--max_old_space_size`, for example to increase the old space to 4GB do:

```shell
$ export NODE_OPTIONS="--max_old_space_size=4096"
```

```

PR #315, opened 2020-04-04
Fix drop-down menu overlap issue on small screens

rimshach:
On smaller screens ,the dropdown menu headings overlapped each other,making navigation difficult.
image

Now this looks like this 
image

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -135,7 +135,7 @@ body {
 }
 
 .navbar-dropdown {
-  position: absolute;
+  
   display: none;
   top: calc(100% - 1px);
   background-color: var(--navbar-menu-background);
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -135,7 +135,6 @@ body {
 }
 
 .navbar-dropdown {
-  
   display: none;
   top: calc(100% - 1px);
   background-color: var(--navbar-menu-background);


aashnajena:
@zregvart I think when you made your PR #306, you didn't realise that you're putting common CSS for all media screens. I don't think the menu on the small screen should be having the box outline, or the hover effect - those were only for >1024px screens... I think we'll need to bring back the CSS for small media screens.

zregvart:
This is what we have currently at 1024px:

Firefox/Chromium:
Screenshot_2020-04-06 Home - Apache Camel

camel apache org_

This is what we have with this PR (Firefox/Chromium:
Screenshot_2020-04-06 Home - Apache Camel(1)
deploy-preview-315--camel netlify com_

At 1023px and lower, the top level navigation is removed and hamburger menu on the top right is shown.

Not sure what the issue is or how this solves it.

aashnajena:
The issue seems to be with mobile menus. Here's how it looks on the current website on my phone :

WhatsApp Image 2020-04-06 at 2 23 50 PM

zregvart:
Ah, okay, now I see. So we need a solution here for both cases, this one breaks the >1024px width case.

rimshach:
If this is already been done as @aashnajena mentioned and css was not merged then i'll close the pull request or if this requires css for <1023 and this is how it should be fixed then i'll do that.

aashnajena:
@rimshach I have actually worked on this menu before so I think I understand exactly where the problem is. I've made some changes at #328 which should fix the issues for all screen sizes. You can test them and include them here if you want..

rimshach:
Oh I did'nt see your pull request before but if you have done then that's great,Thankyou :)No it's fine let's go with the PR you created.I'll close this one.

PR #319, opened 2020-04-05
Fix redirecting to anchor tags

aashnajena:
If we redirect to URLs with anchor tags, the upper portion gets hidden behind the header nav-bar.
Example, try redirecting from https://camel.apache.org/blog/GSoC-2020-announced/ to one of the camel-projects https://camel.apache.org/projects/#apache-camel-k or https://camel.apache.org/projects/#apache-camel-extensions-for-quarkus. The heading is not visible.

As far as my understanding goes, the fragment-jumper file aims at doing this task only. I'm not completely sure about what else the code in the file does, so I haven't deleted it, but the code snippet I have added will take care of all anchor redirections.

### DIFF
--- a/antora-ui-camel/src/css/base.css
+++ b/antora-ui-camel/src/css/base.css
@@ -3,6 +3,7 @@ body {
   height: 100%;
   display: flex;
   flex-direction: column;
+  scroll-padding-top: var(--body-top);
 }
 
 *,
--- a/antora-ui-camel/src/css/vars.css
+++ b/antora-ui-camel/src/css/vars.css
@@ -125,8 +125,8 @@
   --navbar-height: calc(73 / var(--rem-base) * 1rem);
   --toolbar-height: calc(45 / var(--rem-base) * 1rem);
   --drawer-height: var(--toolbar-height);
-  --body-top: var(--navbar-height);
-  --body-min-height: calc(100vh - var(--body-top));
+  --body-top: calc(var(--navbar-height) + 1rem);
+  --body-min-height: calc(100vh - var(--navbar-height));
   --nav-height: calc(var(--body-min-height) - var(--toolbar-height));
   --nav-height--desktop: var(--body-min-height);
   --nav-panel-height: calc(var(--nav-height) - var(--drawer-height));


antora-ui-camel/src/js/02-fragment-jumper.js
```diff
@@ -1,6 +1,10 @@
 ;(function () {
   'use strict'
 
+  var shiftWindow = function () { window.scrollBy(0, -80) }
+  if (window.location.hash) shiftWindow()
```
Delawen, line 5: Nice trick :)

antora-ui-camel/src/js/02-fragment-jumper.js
```diff
@@ -1,6 +1,10 @@
 ;(function () {
   'use strict'
 
+  var shiftWindow = function () { window.scrollBy(0, -80) }
```
Delawen, line 4: These 80 pixels I understand they depend on the size of the bar. Is there no variable we can use here? Thinking on the future, when the bar size may change and this may not adjust properly anymore. I don't like having a long calculation each time this gets called, but it would be good to have some dependency on the height of the nav-bar.

Also, add some comments on the source code explaining what and why this. In a couple of months we will not remember why this was here.

aashnajena: Sure, makes sense. I'm thinking I can find the height of the navbar and add some buffer to it for scrolling. Shouldn't be a long calculation.

Also, can you confirm if the rest of the code in the file is needed? I think it was meant to do this job, but I'm not sure if it serves any other purpose as well. Thanks!

aashnajena:
I hope the changes are better now!

zregvart:
I think this problem should be solved without JavaScript, i.e. with CSS only.

zregvart:
[Here is](https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/) one approach to solving this with CSS.

aashnajena:
@zregvart thanks for sharing, I have made the required changes. The `--body-top` variable was not being used anywhere else, so I altered it to use here.

zregvart:
Thanks!

PR #320, opened 2020-04-05
Camel 14781 security page breadcrumps

Siyubu:


Siyubu:
While navigating on apache website I saw pages that don't have toolbars breadcrumps such as security page from About/ security. 
Problem resolution:
From the root repository under layouts/ partials/seo/structured I built a breadcrumbs.html partial which get and display the current page user is on.
The partial has been included in root/security/list.html unde main div.
The purpose of making breadcamps.html a partial is making easy the act of displaying breadcrumps on any camel-website page. what is needed is to include it in the targeted page.

PR #326, opened 2020-04-06
fixed release pages responsive issue

jyotiattri:
In this PR, I have fixed the release pages responsive issue in mobile view. I have noticed that in release pages some of the text is being cut from the left side, so I have fixed it.
**Before:**
release-before
**After:**
release-after


zregvart:
Thanks!

PR #327, opened 2020-04-06
Search-Cancel Button position fixes in the search box

psingh587:
With the recent changes, the search-cancel button had got disoriented. I made changes to fix it.

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -417,7 +417,7 @@ body {
 #search-cancel {
   position: relative;
   bottom: 1.4rem;
-  left: 10.6rem;
+  left: 7.5rem;
   height: 1.1rem;
   display: none;
   cursor: pointer;
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -416,9 +416,9 @@ body {
 
 #search-cancel {
   position: relative;
-  bottom: 1.4rem;
-  left: 7.5rem;
-  height: 1.1rem;
+  bottom: calc(50% - var(--body-font-size--desktop) / 7.5);
+  left: calc(100% - var(--body-font-size--desktop) / 0.9);
+  height: 1rem;
   display: none;
   cursor: pointer;
 }
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -416,8 +416,8 @@ body {
 
 #search-cancel {
   position: relative;
-  bottom: calc(50% - var(--body-font-size--desktop) / 7.5);
-  left: calc(100% - var(--body-font-size--desktop) / 0.9);
+  bottom: calc(50% - 0.15rem);
+  left: calc(100% - 1.25rem);
   height: 1rem;
   display: none;
   cursor: pointer;


antora-ui-camel/src/css/header.css
```diff
@@ -417,7 +417,7 @@ body {
 #search-cancel {
   position: relative;
   bottom: 1.4rem;
-  left: 10.6rem;
+  left: 7.5rem;
   height: 1.1rem;
```
zregvart, line 421: Perhaps instead of relying on specific numbers we can rely on the principles, this way if we change the search input field size the clear icon will be positioned correctly.

```suggestion
  bottom: calc(50% - 0.15rem);
  left: calc(100% - 1.25rem);
  height: 1rem;
```

Not sure if 0.15 and 1.15 could be derived from a variable, that would be better.

psingh587:
@zregvart  Made the changes.

antora-ui-camel/src/css/header.css
```diff
@@ -416,9 +416,9 @@ body {
 
 #search-cancel {
   position: relative;
-  bottom: 1.4rem;
-  left: 10.6rem;
-  height: 1.1rem;
+  bottom: calc(50% - var(--body-font-size--desktop) / 7.5);
+  left: calc(100% - var(--body-font-size--desktop) / 0.9);
```
zregvart, line 420: What's the significance of 7.5 and 0.9?

psingh587: Actually to convert the existing variable quantity into 0.15 rem and 1.25 rem respectively, i did that. I thought that's what you meant by 'deriving it from a variable'

psingh587: --body-font-size--desktop = 1.125rem here. To make it 0.15 rem and 1.25 rem i divided it with 7.5 and 0.9 respectively. 
I am sorry if I understood it wrongly

zregvart: I thought that we don't put fixed values but reference a variable there (if possible). These are [magic numbers](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) and if we don't have a principle behind those, it's better to have one: a way to correlate a change in one variable to a change in another. So if one changes we don't have to look for other bits that need to change and we don't introduce new issues. For example, and I'm not 100% on this, perhaps these can be derived from `--doc-margin--desktop`?

psingh587: @zregvart I understood about the correlation and the intuition behind it. However, it looks I am finding a hard time deriving it from the variable. Could you please help me here, so that I would remember it in future?

zregvart: @psingh587 could be that there isn't one that we explicitly have captured in a variable. I _think_ we could compute this from the height the icon subtracted from of the input box and then divide that in half to get equal parts above and below the icon. We could also play with padding/margins so we position at 50% and padding/margin makes sure it's centred. There's no need to make this more complicated than it is though. If you're okay with how it is right now I think we can merge this as it does fix a problem we have.

psingh587: If this works, let us merge it.

zregvart:
Thanks!

PR #328, opened 2020-04-06
Fix the nav-dropdown for small screens

aashnajena:
The problem in drop-down menus for small screens :

WhatsApp Image 2020-04-06 at 2 23 50 PM

I think the changes I have made should suffice for now. Menu for small screens looks like this :

small-menu

I feel that the mobile-menu could do with some improvement on the design front, but I want to work on that in a different PR.


zregvart:
Thanks!

PR #329, opened 2020-04-06
Make a smooth scrolling effect for back to top

AemieJ:
* On click of the **Back to top**, the scroll to the top of the page is not smooth. Hence, I made changes so that the scrolling to the top of the page is smooth.

### DIFF
--- a/antora-ui-camel/src/css/base.css
+++ b/antora-ui-camel/src/css/base.css
@@ -3,6 +3,7 @@ body {
   height: 100%;
   display: flex;
   flex-direction: column;
+  scroll-behavior: smooth;
 }
 
 *,


psingh587:
Hi, actually I had suggested putting the same effect. But @zregvart mentioned an article which said that the smooth scrolling sometimes provides a dizzy effect for the end user.
Check the last comment in PR #234.

AemieJ:
> Hi, actually I had suggested putting the same effect. But @zregvart mentioned an article which said that the smooth scrolling sometimes provides a dizzy effect for the end user.
> Check the last comment in PR #234.

@zregvart and @psingh587, sorry about suggesting the same changes but according to my opinion,  when you put a **Back to top** button, to put a smooth or any animation to scroll in upward direction gives better feel than the current one which just jumps to the top page without any effect.

psingh587:
Yeah.. but as i said, in pages with less content, its looks alright. But for pages with larger contents, a lot of them has to be scrolled down in order to reach to the top. And this does gives a dizzy effect. Thats what @zregvart shared and I think it actually is better without a smooth scrolling effect. At the end , user experience matters.

zregvart:
I don't think we want to do this, it can [make people sick](https://css-tricks.com/smooth-scrolling-accessibility/#article-header-id-5). I'd like for us to be more inclusive towards folk with disabilities or ailments. Enabling smooth scrolling also affects battery life. As far as I understand the user can choose to enable this by configuring smooth scrolling in their browser settings.

PR #334, opened 2020-04-07
Fixed table first column font size bold

jyotiattri:
In this PR, I have changed the design of all table that is present on the website. I have seen some tables the first column is bold or some not. I have changed the first column font size bold to make same UI for all the tables.
**Before:**
table-before
**After:**
table-after

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -667,3 +667,7 @@ table.tableblock thead {
   background: var(--color-smoke-90);
   font-weight: var(--body-font-weight-bold);
 }
+
+.doc table.tableblock td:nth-child(1) {
+  font-weight: var(--body-font-weight-bold);
+}
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -671,3 +671,7 @@ table.tableblock thead {
 .doc table.tableblock td:nth-child(1) {
   font-weight: var(--body-font-weight-bold);
 }
+
+.doc .sect1 table.tableblock td:nth-child(2) {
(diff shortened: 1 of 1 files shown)


djencks:
I'm not sure about this either.  Perhaps something similar could be done by making the first column into a header column? See https://asciidoctor.org/docs/user-manual/#cols-format.

jyotiattri:
@zregvart, In EIP documentation we can use there main class and make is same as other tables. In the recent changes, I have made changes in EIP documentation.

jyotiattri:
@djencks, Thanks

zregvart:
I don't think this is the right approach, it will break when we change what text is presented in which column. Better approach is the one that @djencks noted, to include such formatting in the source Asciidoc files. Most tables are already generated and what we're working on improving that. It would be much better, in my opinion, to make changes like this in the code generation bits so that these styling options are applied in the source files.

For the components we're using the `indexTable`, e.g. now:

https://github.com/apache/camel/blob/d02f923ecc08e5515a6742485a6c4fe05aac7416/docs/components/modules/ROOT/pages/index.adoc#L18

I think it would be a good idea to expand the `indexTable` so we can include column formatting options. The source for `indexTable` processor is [here](https://gitlab.com/djencks/asciidoctor-antora-indexer), perhaps it would be best to continue this there and improve on this when it's ready to use in Camel.

jyotiattri:
@zregvart, Thanks for the reply. I'll look into [indexTable](https://gitlab.com/djencks/asciidoctor-antora-indexer) and will update you if I can make changes in it.

zregvart:
I think we can close this now.

jyotiattri:
Yes, please

PR #336, opened 2020-04-07
separator part

raashika03:
@zregvart  please check I'll work on animation effect for the front page on other pr(it's quite uneasy for me to work locally changing so many codes). please check and tell if the changes are ok for the lines or need to change anything. closed the previous pr because commited some mistakes

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -15,7 +15,6 @@ body {
   top: 0;
   width: 100%;
   word-wrap: break-word;
-  z-index: var(--z-index-navbar);
   border-bottom: 1px solid var(--color-smoke-90);
   box-shadow: 0 0.5rem 3rem var(--color-smoke-70);
 }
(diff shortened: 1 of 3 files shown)


zregvart:
@raashika03 you can add more commits and push them to your fork and this pull request will be updated.

We don't have previews so it's a bit difficult to show this. For me the line looks too thick and the animation is a bit too erratic. Perhaps this wasn't the best idea to have the lines animated. I guess what I meant when I created [the issue](https://issues.apache.org/jira/browse/CAMEL-14677) was to have a transition from the swoosh to the line, not to have the line itself animated, at least not in such a profound way. I understand you put a lot of effort into this and that the task is not defined as clearly, but I'm not sure this makes the website better. I'm hoping that we can find a way to incorporate the swoosh, it could also be that this is not a good idea to begin with.

Peek 2020-04-07 15-59

raashika03:
I'll decrease the thickness and will remove the animation effect from the lines and about that transition from swoosh I'm working on that using js will open other pr for that. and these lines will be on all the pages other than the frontpage. and on the frontpage I'll make the transition as per your suggestion @zregvart

zregvart:
@raashika03 there's an interesting technique in [this](https://css-tricks.com/scroll-drawing/) article. Not sure if it's 100% applicable but it can serve as an inspiration.

raashika03:
> @raashika03 there's an interesting technique in [this](https://css-tricks.com/scroll-drawing/) article. Not sure if it's 100% applicable but it can serve as an inspiration.

I went through all these already :) I'm working on it  thank you @zregvart

raashika03:
@zregvart  please check!

zregvart:
Not sure, for my taste it's a bit too much, looks a bit detached from the rest of the design. The CAMEL-14677 was meant as a exploration of ideas on how to incorporate the swoosh thought the website and the lines below the top level navigation was one of those ideas. Not all ideas are perfect.

If you look at the trends in web design, they tend to simplify and use flat styling (e.g. Google Material, Microsoft Fluent).

zregvart:
Thanks for the effort @raashika03. In the end we decided not to proceed with this idea.

PR #338, opened 2020-04-07
Camel 14857 change nav panel explore color

Siyubu:
change nav panel explore color


Siyubu:
Conflict with base branch resolved

PR #340, opened 2020-04-08
Fix issues with search results

aashnajena:
I noticed a couple of issues :

1) "Nothing found" box is very cramped.
no_results_old

Fix:
no_results_new

2) I noticed there's excess space in some search results. For example, try searching for "Book", or "Camel" or "Article". I realised that there were some empty search results being displayed here. As shown in screenshots below, the URL for them exists, but heading & summary doesn't, because these are actually "section names" which match with the query. 

search_results_fix

search_example_2

Fix : I thought of two ways we can go about it - either we don't display these empty results, or we show the section names in their place. Second one seemed apt to me, since all such "missing content" cases that I could find were links to sections. Maybe we can attach links to the section headings and do away with the empty results - but I still thought that if the section-name is matching the query, it should be displayed separately.

search-fix
 
 Please let me know if we would prefer to fix this in some other way.

### DIFF
--- a/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
+++ b/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
@@ -46,7 +46,12 @@
           const data = hits.reduce((data, hit) => {
             const d = {}
             d.url = hit.url
-            d.breadcrumbs = Object.values(hit.hierarchy).slice(1).filter((lvl) => lvl !== null).join(' &raquo; ')
+            var breadcrumbs = Object.values(hit.hierarchy).slice(1).filter((lvl) => lvl !== null).join(' &raquo; ')
+            if (breadcrumbs !== '') {
+              d.breadcrumbs = breadcrumbs
+            } else {
+              d.breadcrumbs = hit.hierarchy.lvl0
+            }
             if (hit._snippetResult !== undefined) {
               d.snippet = hit._snippetResult.content.value
             } else {
@@ -65,7 +70,7 @@
         .then((data) => {
           if (Object.entries(data).length === 0 && data.constructor === Object) {
             return `
-              <header>Nothing Found</header>
+              <header class="no_search_results">Nothing Found</header>
               `
           } else {
             return `
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -462,6 +462,11 @@ body {
   padding-top: 0.3rem;
 }
 
+.no_search_results {
+  padding: 0.5rem 1rem;
+  font-weight: bold;
+}
+
 .algolia-docsearch-suggestion--highlight {
   font-weight: bold;
   color: var(--color-jet-50);


antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -46,7 +46,12 @@
           const data = hits.reduce((data, hit) => {
             const d = {}
             d.url = hit.url
-            d.breadcrumbs = Object.values(hit.hierarchy).slice(1).filter((lvl) => lvl !== null).join(' &raquo; ')
+            var breadcrumbs = Object.values(hit.hierarchy).slice(1).filter((lvl) => lvl !== null).join(' &raquo; ')
```
zregvart, line 49: Do you think it makes sense to remove the `slice(1)`, i.e. start from level 0 on the hierarchy? Would that cover both cases?

aashnajena: Ahh yes makes sense. For some reason I was reading `lvl0` as lv-10 so I was confused why we're slicing at 1 when the section is at level 10.. I'll change that!

aashnajena: But the problem with that would be that for each case the section name will get considered right? Every link will then contain the name of the section. Do we want that?

aashnajena: I tried it - it makes of the form Section-Name >> Topic-Name

search_example

aashnajena: Any verdict in this?

zregvart: How does the default Algolia search UI handle this?

aashnajena: There's some extra whitespace shown for the empty results (I think due to the margin/padding of the result_header class). It's visible in one of the screenshots above

zregvart: We're not using the default Algolia search UI, how does the [default UI](https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/js/) perform in this case?

aashnajena: I'm not very sure about how to check that. What I tried doing was test their live demo. Here, if I search for a brand itself, they only show the matching products (but in this example each product has the brand name in its title)
algolia

Also, the idea of displaying number of matches per section seems good to me. Can we include that in our website?

zregvart: Our current UI as a proof of concept that we did in #163/#81, so this is just our first iteration of the UI. At the point we made this Algolia didn't offer a light version of the UI, perhaps the current (v4) incarnation is light enough to be used instead of our POC.

With search we have a number of issues that we need to address, design being just one of them. I welcome all improvements in this area.

aashnajena: I see. I'm reading about Algolia's v4 UI right now, but including this in the website looks like a slightly big task. I'd be happy to take it up, along with other issues related to search! For now, should I put a hold to these changes and close this PR? Or should we make these temporary changes and then rethink the search design in accordance to Algolia's default UI later?

zregvart: What I understood was that the discussion was about including lvl0 and comparing how the default Algolia UI handles similar case.

I still think this PR improves the search, so we can merge it.

There is a lot of work on the search left to do. I'll create a JIRA issue to open the discussion on that.

zregvart:
I've created [CAMEL-14952](https://issues.apache.org/jira/browse/CAMEL-14952), we can continue the discussion there.

PR #344, opened 2020-04-14
Fixes CAMEL-14896: set up to build using indexer

djencks:
Add indexer extension

### DIFF
--- a/antora-playbook-xrefcheck.yml
+++ b/antora-playbook-xrefcheck.yml
@@ -6,35 +6,30 @@ site:
 content:
   sources:
 
+# manual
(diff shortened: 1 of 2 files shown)


antora-playbook.yml
```diff
@@ -7,7 +7,7 @@ content:
   sources:
 
     - url: git@github.com:apache/camel.git
-      branches: master
+      branches: issue-14874-generate-index-pages
       start_paths:
```
oscerd, line 11: I think the branch is wrong

djencks: OOPS! Thanks, should be fixed now :-)

PR #347, opened 2020-04-16
feat(build): image optimization

zregvart:
This adds image optimization using imageoptim. According to the output
this saves 5.49 MB or 19%. The tradeoff is that the image optimization
takes a long time to perform and would increase the build times, on my
machine ~3:30min more.


zregvart:
There is an issue with svgo, it seems to clober the brand-logos.svg

zregvart:
There seems to be [an issue](https://builds.apache.org/job/Camel.website/job/pr%252Fimage-optimization/4/) with processing JPEG files. I've seen this on my computer as well that's why I've pre-optimized the `camel-featured.jpg` files. The issue seems to occur intermittently in the website build on Jenkins as well.

zregvart:
For some reason every other build fails. Moving this to draft so it doesn't get merged by mistake.

AemieJ:
@zregvart I wanted to talk regarding this as the images tend to reduce the performance on each blog. Are we considering using a Github app such as imgbot to continuously optimize the images or such?

zregvart:
@AemieJ for imgbot we need to reachout to ASF INFRA to enable/install it on our repository, is there a github action we can use? Perhaps I should rebase this and see if there's a newer version of imagemin that doesn't crash sporadically.

AemieJ:
> @AemieJ for imgbot we need to reachout to ASF INFRA to enable/install it on our repository, is there a github action we can use? Perhaps I should rebase this and see if there's a newer version of imagemin that doesn't crash sporadically.

I did research on the GitHub action for compressing the image, I found a quite popular one https://github.com/marketplace/actions/image-actions. I think we could give this a try if you haven't given it a try previously?

djencks:
I think the Antora UI build uses a gulp extension to optimize images.  Perhaps it would be worthwhile investigating and possibly using the same mechanism?

David Jencks

zregvart:
Antora UI uses gulp-imagemin (version 6.2), we're trying to use the same here (version 7), we experience segmentation faults with this.

PR #349, opened 2020-04-19
Change right padding for header menu-items with drop-downs

aashnajena:
In the header menu, the spacing between Community and Download is less than others, due to the space taken up by the `::after` drop-down arrow.    

image

I've added extra right padding for menu-items having drop-downs so that spacing seems uniform

image


zregvart:
Thanks!

PR #353, opened 2020-04-24
Change category-wise blog layout and format the code within blog

AemieJ:
* Within the blog layout, when a user clicked on a particular category it would move to a category-wise based page. However, if the user wanted to check from some other category, it would be difficult, thus reducing the accessibility and easiness of reading the blogs. 

* Hence, I thought that for each category-wise based page, all the categories must be included to so easy navigation can be performed thus increasing the accessibility for different blogs. 

* The code wasn't formatted, hence I made a few changes to that as well.

### DIFF
--- a/layouts/_default/taxonomy.html
+++ b/layouts/_default/taxonomy.html
@@ -1,10 +1,28 @@
 {{ partial "header.html" . }}
 
 <div class="body">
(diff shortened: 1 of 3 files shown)


zregvart:
Thanks!

PR #355, opened 2020-04-27
Change break-all to break-word in release notes

aashnajena:
In the tables present on release pages, the words should not break when screen size decreases (see the right column). This makes it harder to read the information. The break should be at word boundaries.
 
image


zregvart:
Thanks!

PR #360, opened 2020-05-01
Add logos on the projects page

avgupt:
Added logos on the projects page as suggested [here](https://issues.apache.org/jira/browse/CAMEL-14994)


zregvart:
Work in #370 superseded this.

PR #366, opened 2020-05-19
feat: change frontpage swoosh animation

zregvart:
To freshen up the frontpage a bit, and perhaps to be more representative
of the integration nature of Camel, this changes the swoosh animation to
a "transmission" animation. By using dash pattern of the same SVG paths
we can simulate "messages" traveling from left and right edges of the
screen.

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -84,13 +84,15 @@ header.frontpage svg {
 
 header.frontpage svg path {
   fill: none;
-  animation: sparkle 2s ease alternate infinite;
+  stroke-linecap: round;
+  animation: transmission 10s ease alternate infinite;
 }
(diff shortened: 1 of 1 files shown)


zregvart:
Preview is [here](https://5ec42c32be4a7300063576f2--camel.netlify.app/).

AemieJ:
> Preview is [here](https://5ec42c32be4a7300063576f2--camel.netlify.app/).

In my opinion, we could make the animation a bit slow and components within an animation spread out 'cause they are really close and timing per frame is too fast. It creates dizziness after a while. It would be more soothing if it were a slower time per frame.

aashnajena:
Looks really good to me!

davsclaus:
The front page has been redesigned so the swoosh animation is no long there, so lets close this PR

PR #368, opened 2020-05-22
Status badges CSS for Camel Quarkus extension pages

ppalaga:
This allows to have badges for since version, JVM and Native on Camel Quarkus Extension pages, like the following:
image

Can be tested together with https://github.com/apache/camel-quarkus/pull/1260

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -667,3 +667,36 @@ table.tableblock thead {
   background: var(--color-smoke-90);
   font-weight: var(--body-font-weight-bold);
 }
+
+/* Status badges used in Camel Quarkus extension pages */
+div.badges p {
+  text-align: right;
+}
+
+div.badges span {
+  color: var(--color-white);
+  padding: 0.15rem 0.3rem 0.15rem 0.3rem;
+  font-weight: normal;
+  font-style: normal;
+  font-size: 0.8rem;
+}
+
+div.badges span.badge-key {
+  background-color: var(--color-gray-60);
+  border-radius: 0.2rem 0 0 0.2rem;
+}
+
+div.badges span.badge-version {
+  background-color: var(--note-color);
+  border-radius: 0 0.2rem 0.2rem 0;
+}
+
+div.badges span.badge-supported {
+  background-color: var(--tip-color);
+  border-radius: 0 0.2rem 0.2rem 0;
+}
+
+div.badges span.badge-unsupported {
+  background-color: var(--important-color);
+  border-radius: 0 0.2rem 0.2rem 0;
+}


ppalaga:
Do I need to care for the failed CI? The failure messages do not look very informative.

zregvart:
@ppalaga I've restarted the build without cache and that seems to have helped preview is [here](https://deploy-preview-368--camel.netlify.app/camel-quarkus/latest/extensions/activemq.html), though I'm guessing there is no markup present in the camel-quarkus documentation at the moment to take advantage of the css.

antora-ui-camel/src/css/doc.css
```diff
@@ -667,3 +667,36 @@ table.tableblock thead {
...
 }
+
+/* Status badges used in Camel Quarkus extension pages */
+div.badges p {
+  text-align: right;
+}
+
+div.badges span {
+  color: var(--color-white);
+  padding: 2px 6px 2px 6px;
```
zregvart, line 678: Prefer using relative values, e.g. `rem`, test by changing the base font size to see if the proportions work. This is to help folk that configure their browsers with larger/smaller default text sizes. This works for almost any usage, though sometimes using `px` makes sense, like for borders.

antora-ui-camel/src/css/doc.css
```diff
@@ -667,3 +667,36 @@ table.tableblock thead {
...
+  font-weight: normal;
+  font-style: normal;
+  font-size: 0.8em;
```
zregvart, line 681:
```suggestion
  font-size: 0.8rem;
```

antora-ui-camel/src/css/doc.css
ppalaga: Thanks, done in d140a11

antora-ui-camel/src/css/doc.css
ppalaga: Done in d140a11

ppalaga:
> I'm guessing there is no markup present in the camel-quarkus documentation at the moment

Yes, apache/camel-quarkus#1260 is blocked by the current PR :)

I have tested d140a11 locally with various default font sizes in Chrome and all of them worked as expected.

ppalaga:
> re-use the badge of since event for the camel componets.

Producer/consumer support is another good candidate. 

Perhaps some UI expert can re-style these supported/unsupported badges to something less talkative, maybe some sort of ✔/✘ instead of `suṕported/unsupported`

ppalaga:
The CI failed again. I'd thankful for any suggestion what to do.

oscerd:
I think it's not totally reliable.. right @zregvart ?

ppalaga:
Merged, as agreed with @oscerd . Let's revert in case there is any damage.

PR #271, opened 2020-03-25
Fixes Wastage of screen real-estate on large displays

avgupt:
Fixes: https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-14769?filter=allopenissues

PR #369, opened 2020-05-23
chore: customized scrollbar for table and code snippets

AemieJ:
* The default scrollbar didn't amount to the design of the webpage, hence I added a custom scrollbar using WebKit for all the table layouts and code snippets.

### Table Layout Before 
table-scrollbar-before

### Table Layout After 
table-scrollbar-after
<hr />

### Code Layout Before 
code-context-before

### Code Layout After
code-context-after

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -111,6 +111,25 @@
   overflow-x: auto;
 }
 
+.doc .table-wrapper::-webkit-scrollbar {
+  height: 8px;
+}
+
+.doc .table-wrapper::-webkit-scrollbar-track {
+  background: #c4c4c4;
+  border-radius: 50px;
+}
+
+.doc .table-wrapper::-webkit-scrollbar-thumb {
+  background: rgb(151, 151, 151);
+  border-radius: 50px;
+}
+
+.doc .table-wrapper::-webkit-scrollbar-thumb:hover,
+.doc .table-wrapper::-webkit-scrollbar-thumb:active {
+  background: rgb(107, 107, 107);
+}
+
 .doc .table-wrapper table {
   width: 100%;
 }
@@ -530,6 +549,25 @@
   display: block;
   overflow-x: auto;
   padding: 0.75rem;
+}
+
+.doc pre.highlight code::-webkit-scrollbar {
+  height: 8px;
+}
+
+.doc pre.highlight code::-webkit-scrollbar-track {
+  background: #c4c4c4;
+  border-radius: 50px;
+}
+
+.doc pre.highlight code::-webkit-scrollbar-thumb {
+  background: rgb(151, 151, 151);
+  border-radius: 50px;
+}
+
+.doc pre.highlight code::-webkit-scrollbar-thumb:hover,
+.doc pre.highlight code::-webkit-scrollbar-thumb:active {
+  background: rgb(107, 107, 107);
 }
 
 /* NOTE assume pre.highlight contains code[data-lang] */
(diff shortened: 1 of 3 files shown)


zregvart:
I would prefer if we don't target specific browser, this makes it hard for us to maintain in the future (we need to test those rules in many browsers/versions).

I think this doesn't help much with the way we display tables (they're too wide), there has been some work in #271, I think it would be better to finish that work so by default we don't have horizontal scrollbars.

AemieJ:
> I would prefer if we don't target specific browser, this makes it hard for us to maintain in the future (we need to test those rules in many browsers/versions).
> 
> I think this doesn't help much with the way we display tables (they're too wide), there has been some work in #271, I think it would be better to finish that work so by default we don't have horizontal scrollbars.

Yes, if the table are going to be prevented from having overflow, that's a real good thing. Does the same thing resolve for the code snippets?

AemieJ:
Alright, so you want to keep the current design for scrollbars? If so, I will close this without any further changes to it.

zregvart:
@AemieJ let's discuss this on the call on Monday.

zregvart:
On today's call we agreed to research the accessibility implications of this change.

AemieJ:
@zregvart I will do the research on it and get back to you

AemieJ:
@zregvart I researched on a couple of websites for accessibility implications of custom scrollbars. Based on it, I came to the conclusion that the contrast of the scrollbar track with the page should be 4.5:1 while that of scrollbar thumb and track to be 3:1 for proper contrast ratio. Other concerns on the width of the scrollbar, yes it can be adjusted through the different OS but it tends to fail in few cases so a minimum of 48px is recommended for the width of the scrollbar and not much as we require more of the content to be observed. Also, we keep the default obstructive layout with designs of scrollbar similar to macOS that is focused on aesthetic styling. 

Another issue I wanted to raise and need opinions in is do we need consistency in the scrollbar in the sense similar scrollbar styling for cross browsers or not?

AemieJ:
@zregvart one can always control the width of scrollbars, I checked and one of the reasons I wanted to make use of custom scrollbars is to improve its aesthetics however I need to increase the width and improve color contrast compared to the default.

AemieJ:
@zregvart In addition to what I observed from the camel website is that for smaller screen width, the scrollbar has an unobstructed layout and if it needs to be easily accessible I believe, for any screen width, we need to include a scrollbar with an obstructive layout.

zregvart:
This needs to be rebased.

AemieJ:
@zregvart I have done that now!

antora-ui-camel/src/css/doc.css
```diff
@@ -111,6 +111,25 @@
   overflow-x: auto;
 }
 
+.doc .table-wrapper::-webkit-scrollbar {
+  height: 0.42rem;
```
zregvart, line 115: Perhaps this and the 3.125 value can be put in the var.css as well.

antora-ui-camel/src/css/vars.css
```diff
@@ -150,6 +150,8 @@
   --doc-max-width--desktop: calc(1366 / var(--rem-base) * 1rem);
   --static-max-width: calc(720 / var(--rem-base) * 1rem);
   --static-max-width--desktop: calc(1366 / var(--rem-base) * 1rem);
+  --scrollbar-thickness: calc(7.56 / var(--rem-base) * 1rem);
+  --scrollbar-radius: calc(56.25 / var(--rem-base) * 1rem);
```
zregvart, line 154: I'd move this with the other `--scrollbar` variables. This is somewhat confusing calculation, perhaps use the `em` like in `--body-font-size*` and add a comment as to how many pixels that corresponds to.

zregvart:
Thanks!

PR #377, opened 2020-05-29
Add script to remove over lapping of nav-panel menu with context

AemieJ:
1. Less than 1024 px, when clicked on the nav-toggle, it overlaps the context and doesn't look sophisticated. Hence, I made the following changes: 

* 975px <= width < 1024px: When the `nav-toggle is-active` is observed, the context moves to provide a clear line between the nav panel and the context. When the state is back to `nav-toggle`, the default `margin-right: auto` appears. 

* width < 975px: For such less screen width, on click of the **nav-toggle**, the **nav-panel-menu** covers the width of the screen.

### Suggestion
In width > 1024px, would it be better to include a box-shadow beneath the quick lookup similar to that of the toolbar?

### DIFF
--- a/antora-ui-camel/src/js/06-nav-toggle.js
+++ b/antora-ui-camel/src/js/06-nav-toggle.js
@@ -1,0 +1,39 @@
+;(function () {
+  'use strict'
+
+  var MutationObserver = window.MutationObserver
+  var width = window.innerWidth
+  var navToggle = document.getElementsByClassName('nav-toggle')[0]
+  var toggleObserver = new MutationObserver(toggleWidthChange)
+  var config = {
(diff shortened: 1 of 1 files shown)


zregvart:
We have merged #379, is this still needed?

AemieJ:
> We have merged #379, is this still needed?

This is a different issue compared to that one. In this in smaller screen width, the paragraph Doc is overlapped by the menu panel.

zregvart:
I'm not clear on the problem this tries to address. I would prefer a CSS solution rather than JavaScript. @AemieJ it would be helpful if you can provide a screenshot of the issue.

PR #378, opened 2020-05-30
CAMEL-14904: Nav pane does not show current page

AemieJ:
### Issue Observed
* The Antora default UI highlights and makes sure the nav list item for the current page remains visible.  The Camel components list does not do this, always scrolling to the top when you click on an item.  This makes it very hard to use.

### Solution Offered 
* As the `panel.scrollTop` has been used to calculate the mid-point for the `current-page-item` on the nav menu panel, I have written code such that when the navigation happens, it find the midpoint distance for that page item in the menu panel, and scrolls to that immediately.

### DIFF
--- a/antora-ui-camel/src/js/01-nav.js
+++ b/antora-ui-camel/src/js/01-nav.js
@@ -91,6 +91,9 @@
     var navStyle = window.getComputedStyle(nav)
     if (navStyle.position === 'sticky') effectiveHeight -= (rect.top - parseFloat(navStyle.top))
     panel.scrollTop = Math.max(0, (el.getBoundingClientRect().height - effectiveHeight) * 0.5 + el.offsetTop)
+    var topPixels = Math.max(0, (el.getBoundingClientRect().height - effectiveHeight) * 0.5 + el.offsetTop)
+    var navMenu = document.getElementsByClassName('nav-menu')[0]
+    navMenu.scrollBy(0, topPixels)
   }
 
   function find (from, selector) {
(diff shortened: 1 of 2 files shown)


djencks:
Is this the same technique that the Antora default UI uses?  I think keeping the camel UI as close as possible to the Antora default is extremely important.  Other people may have other ideas.

AemieJ:
@djencks No using the panel.scrollTop calculates the pixels however inspite the similar code on each navigation it doesn't scroll to midpoint. Hence, I just added the function to do the following. It uses the same functions and resolves the issue.

AemieJ:
@zregvart yes, within the scroll to midpoint function, it calculates the midpoint using the parent element similar to the antora ui builder and just using the scrollBy function can be used to use the actual pixel values calculated, if we want to avoid rebasing. 

It is the same manner done for the antora docs website. The selected page item moves to the midpoint.

zregvart:
@AemieJ perhaps I wasn't too clear, sorry about that, I think that the `panel` passed into `scrollItemToMidpoint` doesn't need to be the `div` with `data-panel="menu"` it could be the `nav` that we do scroll. That is, make the same change we made in CSS in the JavaScript, we don't scroll that `div`, we scroll the `nav`, keeping the quick search fixed at the top. What I think can be done is to move `data-panel="menu"` to the `nav`, I think that then we might not need any JavaScript changes.

AemieJ:
@zregvart I tried out those changes it works perfectly without any change in javascript, just a change in the handlebar.

PR #379, opened 2020-06-01
Replicate Antora styling for overlapping of navbar over menu panel

AemieJ:
### ISSUES
1.  When clicked on the menu panel within smaller screen width and at the same time, the navbar was clicked it created quite a cluster and looked messy. 
issue-mutation

2. When the menu panel was active and when the dropdown button within the navbar was clicked, the menu panel would become inactive and that isn't user-friendly. 

### SOLUTION
1.  Replicated the way an antora performs this. The navbar takes the entire width while covering 50% of viewport height and shows a higher hierarchy over the menu panel item. 
resolve-mutation

2. `e.stopPropagation()` was used after the click of the dropdown item to resolve this issue.

### DIFF
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -160,13 +160,12 @@ body {
   .navbar-menu.is-active {
     display: block;
     position: absolute;
-    height: 100vh;
+    height: 50vh;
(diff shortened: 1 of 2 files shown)


zregvart:
Thanks!

PR #382, opened 2020-06-03
refactor: use year/month in blog post URLs

fvaleri:


oscerd:
Makes sense. Let's close this

fvaleri:
> This will lead to broken links. We can't do this without adding redirects from existing URLs to the new ones.

Right. 

Given that our 404 page URL is like this:
```
https://camel.apache.org/blog/TestZero/
```
We could add something as simple as:
```
var baseUrl = "https://camel.apache.org/blog/";
var redirects = {
    'TestZero': baseUrl + '2019/12/TestZero/',
    'TestOne': baseUrl + '2020/01/TestOne/'
};

var url = window.location.href.replace(/\/$/, '');
var oldPath = url.substr(url.lastIndexOf('/') + 1);

var newPath = redirects[oldPath];
if (newPath) {
    console.log('Redirecting ' + oldPath + ' to '+ newPath);
    window.location.replace(newPath);
}
```

It's just plain JS, but I'm not familiar with the website framework, so I don't know if something like this can be easily added. Such small feature could be useful in other situations too.

fvaleri:
@zregvart Just saw your comment about `static/.htaccess`, will add redirects there.

fvaleri:
@zregvart just tried to set the url property on my last post and run the website locally. It seems to work fine.

If I add `url: /blog/CdcWithCamelAndDebezium` to the `index.md`, then I get:
```
├── 2020
│   ├── 05
│   │   └── CdcWithCamelAndDebezium
│   │       ├── camel-featured.jpg
│   │       ├── camel-featured_hua29054ae124e0a556509e94b698dcf1b_122153_800x0_resize_q95_gaussian.jpg
│   │       └── index.html
├── CdcWithCamelAndDebezium
│   ├── camel-featured.jpg
│   ├── camel-featured_hua29054ae124e0a556509e94b698dcf1b_122153_800x0_resize_q95_gaussian.jpg
│   └── index.html
```

So no redirect would be needed for release posts (they all have url set), while for the other posts we would just need to add the url property. What do you think?

zregvart:
@fvaleri, not sure what would would we gain with organizing the `content/blog` by date and leaving the URLs as is, if anything that might be a bit confusing. Not sure if you cleared the `public` folder before you built, there should not be two copies of the blog post.

I think both the URLs and the organization in the repository need to follow the same convention. So if there's merit in doing that then it needs to be done by adding redirects for existing content and configuring the [permalinks](https://gohugo.io/content-management/urls/#permalinks-configuration-example) for blog posts.

fvaleri:
Hi @zregvart , I think this is a pretty common scheme for organizing blog posts and there seems to be consensus around it. I fully agree with your suggestions, so now there are redirects for all old posts (we already have one published with the new layout) and permalinks configuration.

All links and redirects seem to work fine. Note that the `url` property has been removed in each release post and relative redirect rules has been created. This is to have the same behavior for all posts. If this PR is merged, all future releases should not have `url` property or it should follow the `blog/aaaa/mm/folder-name` layout.

static/.htaccess
```diff
@@ -1053,6 +1053,46 @@ Redirect 301 /zip-file-dataformat.html /components/latest/dataformats/zipfile-da
...
+Redirect /blog/Outreachy-May2020-Update/ /blog/2020/03/Outreachy-May2020-Update/
+Redirect /blog/100Camels/ /blog/2020/04/100Camels/
+Redirect /blog/Camel-Kafka-connector-release-0.1.0/ /blog/2020/04/Camel-Kafka-connector-release-0.1.0/
+Redirect /blog/Camel-Quarkus-release-1.0.0-M7/ /blog/2020/04/Camel-Quarkus-release-1.0.0-M7/
+Redirect /blog/Camel32-Whatsnew/ /blog/2020/04/Camel32-Whatsnew/
+Redirect /blog/release-2-25-1.html /blog/2020/04/RELEASE-2.25.1/
+Redirect /blog/release-3-2-0.html /blog/2020/04/RELEASE-3.2.0/
+Redirect /blog/Camel33-Whatsnew/ /blog/2020/05/Camel33-Whatsnew/
+Redirect /blog/CdcWithCamelAndDebezium/ /blog/2020/05/CdcWithCamelAndDebezium/
+Redirect /blog/release-3-3-0.html /blog/2020/05/RELEASE-3.3.0/
```
zregvart, line 1094: These need to be `Redirect 301` to be permanent redirects

fvaleri: OK

README.md
```diff
@@ -199,9 +199,11 @@ areas of the website:
 
 Use the `blog` archetype to create a new markdown content file in `content/blog`:
 
-    $ yarn hugp new --kind blog blog/URL-SLUG/index.md # replace URL-SLUG with URL slug for the blog post
+    $ yarn hugp new --kind blog blog/POST-HOME/index.md # replace POST-HOME with the pattern aaaa/mm/url-slug
```
zregvart, line 202: Perhaps we can make it a bit more clear:
```suggestion
    $ yarn hugp new --kind blog blog/YYYY/MM/index.md # replace YYYY with the year and MM with the month
```

fvaleri: Thanks. We also need the PostName in the URL scheme in addition to YYYY and MM.

zregvart:
Thanks  @fvaleri!

PR #385, opened 2020-06-07
chore(ci): GitHub action to publish to Netlify

zregvart:
This adds a GitHub action to build and publish to Netlify.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: Build and Deploy to Netlify
+on:
+  push:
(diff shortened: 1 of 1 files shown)


zregvart:
This is not deploying to Netlify because secrets are not passed to pull requests created from forks. We need this exactly for pull requests from forks. There is still a benefit in this as the build and checks would run. To publish to Netlify for our use case we need to find a different solution.

zregvart:
Netlify seems to have fixed the issue, I'm going to close this I don't think we need it.

PR #386, opened 2020-06-08
chore: remove the width toggle

zregvart:
This sets the `max-width` to `auto` which should allow the content to
take up all the available space.


aashnajena:
I'm not very happy with how this makes the documentation/ download/ projects page look. The documentation pages are also looking a bit clustered. I'm not sure if this is the solution for all screen types. On a big screen, like the iMac I was using on call today, this makes more sense, but on my laptop screen I would much rather prefer the first option. Maybe we can switch the default layout to be wide, but I'd still want to at least keep the narrow option present. 

image

image

AemieJ:
@zregvart Why don't we use the width as the same manner as, in antora docs, it looks pretty sophisticated even if we want to include the content on the right-hand side for the large screen width.

no-width-toggle

zregvart:
@aashnajena @AemieJ nice catch, thanks! I wasn't really testing this, I've made a change in 1bc995a77936490e6b51386a115ac7a4883aa15f to base the width on 1366px (~76rem). It's not really targeting 1366px screens as we have navigation pane on the left (and soon on the right), but it leaves some margins on the left and right. The base was 720px (46rem) to me seems too narrow for the tables in the component reference.

Please have a look at the preview and share your feedback.

AemieJ:
@zregvart is the preview in regard to including the context on the right? If so, then the max-width could be reduced. Also, I observed in some screen width, the margin-left isn't taken into account when we could, 'cause when the nav menu panel is present it's best to provide space between the panel menu and doc data. Other than that, it looks better without the wide/narrow toggle.
Also, about the tables, I feel they are all cramped up and the data looks messy in a few tables 'cause there isn't much padding around the columns.

PR #395, opened 2020-06-12
For CAMEL-15182: bump indexer version to support dlists

djencks:
This is the camel-website change for the issue.


djencks:
I don't see any information about what the problem at netlify is, and it looks like almost all our builds there fail with identical minimal information.
Is this PR safe to apply?  It works fine locally.

zregvart:
@djencks see the [discussion](https://lists.apache.org/thread.html/rce07ea051ffe43b0cbd0aedea68c7e6f836e286a88507286efb22687%40%3Cdev.camel.apache.org%3E) on the mailing list for the Netlify issue.

PR #396, opened 2020-06-16
CAMEL-15161: Modify the design for functionality section on frontpage

AemieJ:
* In JIRA Issue 15161, the issue mentioned was that on the front page then the 4 panels for patterns, components, runtimes, data formats take up too much space. 

* Thus, I modified such that the design for the panel is similar to that of the project section on the frontpage, it gives a minimalist and consistent effect and takes less space also. 

### BEFORE
old-functionality

### AFTER
new-functionality

### DIFF
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -201,7 +201,7 @@ section.frontpage h2 {
   flex-direction: column;
   margin: 1rem;
   padding: 1.2rem 0.75rem;
-  border: 1px solid #e1e1e1;
+  border: 1px solid var(--color-smoke-90);
 }
 
 .frontpage .box div {
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -144,7 +144,7 @@ section.frontpage.columns {
   display: flex;
   flex-wrap: wrap;
   border-top: 1px solid var(--color-smoke-50);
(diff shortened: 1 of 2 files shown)


antora-ui-camel/src/css/frontpage.css
```diff
@@ -197,12 +197,14 @@ section.frontpage h2 {
 }
 
 .frontpage .box {
-  flex: 50%;
-  padding: 1.2rem 0;
+  flex: 33%;
+  flex-direction: column;
+  margin: 1rem;
+  padding: 1.2rem 0.75rem;
+  border: 1px solid #e1e1e1;
```
zregvart, line 204: We need to use variables

```suggestion
  border: 1px solid var(--color-smoke-90);
```

antora-ui-camel/src/css/vars.css
```diff
@@ -117,7 +117,7 @@
   /* static */
   --static-margin: 0 auto;
   /* frontpage */
-  --frontpage-max-width: calc(1140 / var(--rem-base) * 1rem);
+  --frontpage-max-width: calc(1350 / var(--rem-base) * 1rem);
```
zregvart, line 120: With the Antora rebase we have 1366px as width, would be good to be consistent here.
```suggestion
  --frontpage-max-width: calc(1366 / var(--rem-base) * 1rem);
```

AemieJ: Applied the changes as per suggestions.

PR #397, opened 2020-06-16
CAMEL-14994: Merge Projects and Documentation Page

aashnajena:
First attempt at simply merging the two pages. We still need to decide if the combined page should be called "Projects" or "Documentation". 

Preview :

Combined_Projects_Page


zregvart:
My preference would be to name this section as "Documentation", and have it under `/docs/`.

antora-ui-camel/src/css/projects.css
```diff
@@ -2,6 +2,21 @@
...
 
+.projects .camel-project {
+  width: 96%;
+  display: inline-flex;
+  border: 1px solid var(--color-smoke-90);
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+  padding: 1rem 1rem 2rem 1rem;
+  margin: 1rem;
```
AemieJ, line 13: I observed from the preview that the alignment is not present below < 1024 and the documentation frame appears shrunk compared to that of project frames. It's best to maintain the same structure, so for below < 1024px width, you should perform the following changes: `width: auto` and `margin: auto`.

antora-ui-camel/src/css/projects.css
```diff
@@ -2,6 +2,21 @@
...
+}
+
+.projects .camel-project .project {
+  border: none;
+}
+
 .projects .project {
   display: inline-flex;
   border: 1px solid var(--color-smoke-90);
```
AemieJ, line 22: Also, for the documentation frame, could you use an additional class so within `.projects .documentation`, you could make the following change - 
```
.projects .documentation {
    width: 45%;
} 

@media screen and (max-width: 1023px) {
  .projects .documentation {
      width: 100%;
  }
}
```
The reason I am suggesting this is 'cause due to `width: 46%` even at 1024px the `user-manual` and `components` get separated and I don't find that the ideal behavior. 

### Before At 1024px
before-frame-1024

### After at 1024 if you could make those changes
after-frame-1024

aashnajena:
@zregvart @AemieJ I have made all the suggested changes wrt content and styling. Please give it another look.

Since we have 5 sub-projects, we have some space left towards the right of Camel Kafka. If there are any ideas about making it look better, please let me know. Not such a big issue though. Whenever a new camel sub-project comes up, we can add it there!

zregvart:
This seems to have interfered with the frontpage:
Screenshot_2020-06-25 Home - Apache Camel(1)

aashnajena:
Scoping css rules to .docs is the problem. I have removed that.

zregvart:
@AemieJ @delawen can you have a look at this and see if you can spot any issues?

Delawen:
I like it

antora-ui-camel/src/css/docs.css
```diff
@@ -1,54 +1,70 @@
-.projects h2 {
+.docs h2 {
   text-align: center;
 }
 
-.projects .project {
+.camel-project {
+  width: calc(100% - 2.5rem);
```
AemieJ, line 6: The width isn't calculated to be in the center of the page horizontally. Best set the width to 
`width: calc(100% - 2.2rem)`

antora-ui-camel/src/css/docs.css
```diff
@@ -1,54 +1,70 @@
...
+}
+
+.camel-project .project {
+  border: none;
+}
+
+.project {
+  display: inline-flex;
+  border: 1px solid var(--color-smoke-90);
+  width: calc(50% - 2.5rem);
```
AemieJ, line 23: The width isn't calculated to be in the center of the page horizontally. Best set the width to
`width: calc(50% - 2.2rem)`

AemieJ:
@aashnajena @zregvart @Delawen design looks good to me. I just found a tiny problem that as width decreases the section of projects and camel-project is not centered horizontally. 

This isn't centered and is more towards the left side. Only change needs to be made in width which I have suggested within the review. 
wrong-width

aashnajena:
Pushed changes for Aemie's comments

AemieJ:
LGTM.

PR #398, opened 2020-06-17
CAMEL-14958: Search Bar Design Implementation

AemieJ:
## DESCRIPTION ON THE PROBLEM

Below the width of 1275px for the screen, the search panel couldn't be included. This is very inconvenient as we require the presence of search for all screen width.

## DESIGN ALTERNATIVE PROVIDED  

The search design has been inspired by the design of https://getbootstrap.com/docs/4.5/getting-started/introduction/. 
The search bar design is present for the width < 1275 where 1275 is the breakpoint. Here, the search bar is present alongside the navbar burger right below the major navbar. Due to the addition of components, changes are made to the nav-panel on various pages and docs as well. 

The hierarchy of this design is as follows: 
```
-- Search Results # If the navbar menu and/or nav panel item are open, the search results will be shown above them
  -- Navbar Menu  # If the nav panel item is open, the navbar menu will appear on top of it
     -- Nav Panel Items  # Least in the hierarchy
```

### BASIC DESIGN IMPLEMENTATION EXAMPLE 
search-bar-design

### DIFF
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -193,7 +193,7 @@ article.blog p {
   padding: 1rem;
 }
 
-@media screen and (min-width: 1175px) {
+@media screen and (min-width: 1024px) {
   .blog .pagination {
     max-width: var(--static-max-width--desktop);
   }
@@ -224,18 +224,12 @@ article.blog p {
   color: var(--color-white);
 }
 
-@media screen and (max-width: 1174px) {
+@media screen and (max-width: 1023px) {
   article.blog:first-child {
     margin-top: 5rem;
   }
 
   .blog.list aside {
     margin-top: 4.5rem;
-  }
-
-  @media screen and (min-width: 1024px) {
-    .blog.list aside {
-      top: var(--navbar-mobile-height);
-    }
   }
 }
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -193,7 +193,7 @@ article.blog p {
   padding: 1rem;
 }
 
-@media screen and (min-width: 1024px) {
+@media screen and (min-width: 1278px) {
   .blog .pagination {
     max-width: var(--static-max-width--desktop);
   }
@@ -222,4 +222,14 @@ article.blog p {
 
 .blog .page-item.active a {
   color: var(--color-white);
+}
+
+@media screen and (max-width: 1277px) {
+  article.blog:first-child {
+    margin-top: 5rem;
+  }
+
+  .blog.list aside {
+    margin-top: 4.5rem;
+  }
 }
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -232,4 +232,10 @@ article.blog p {
   .blog.list aside {
     margin-top: 4.5rem;
   }
+
+  @media screen and (min-width: 1024px) {
+    .blog.list aside {
+      top: var(--navbar-mobile-height);
+    }
+  }
 }
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -193,7 +193,7 @@ article.blog p {
   padding: 1rem;
 }
 
-@media screen and (min-width: 1278px) {
+@media screen and (min-width: 1175px) {
   .blog .pagination {
     max-width: var(--static-max-width--desktop);
   }
@@ -224,7 +224,7 @@ article.blog p {
   color: var(--color-white);
 }
 
-@media screen and (max-width: 1277px) {
+@media screen and (max-width: 1174px) {
   article.blog:first-child {
     margin-top: 5rem;
   }
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -225,10 +225,6 @@ article.blog p {
 }
 
 @media screen and (max-width: 1023px) {
-  article.blog:first-child {
-    margin-top: 5rem;
-  }
-
   .blog.list aside {
     margin-top: 4.5rem;
   }
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -37,21 +37,15 @@
   margin: 1.5rem 0;
 }
 
-@media screen and (min-width: 1278px) {
+@media screen and (min-width: 1175px) {
   .doc > h1.page:first-child {
     margin-top: 2.5rem;
   }
 }
 
-@media screen and (max-width: 1277px) {
+@media screen and (max-width: 1174px) {
   .doc > h1.page:first-child {
-    margin-top: 3rem;
-  }
-
-  @media screen and (max-width: 1023px) {
-    .doc > h1.page:first-child {
-      margin-top: 5rem;
-    }
+    margin-top: 5rem;
   }
 }
 
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -37,9 +37,21 @@
   margin: 1.5rem 0;
 }
 
-@media screen and (min-width: 769px) {
+@media screen and (min-width: 1278px) {
   .doc > h1.page:first-child {
     margin-top: 2.5rem;
+  }
+}
+
+@media screen and (max-width: 1277px) {
+  .doc > h1.page:first-child {
+    margin-top: 3rem;
+  }
+
+  @media screen and (max-width: 1023px) {
+    .doc > h1.page:first-child {
+      margin-top: 5rem;
+    }
   }
 }
 
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -37,18 +37,6 @@
   margin: 1.5rem 0;
 }
 
-@media screen and (min-width: 1024px) {
-  .doc > h1.page:first-child {
-    margin-top: 2.5rem;
-  }
-}
-
-@media screen and (max-width: 1023px) {
-  .doc > h1.page:first-child {
-    margin-top: 5rem;
-  }
-}
-
 #preamble + .sect1,
 .doc .sect1 + .sect1 {
   margin-top: 2rem;
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -37,13 +37,13 @@
   margin: 1.5rem 0;
 }
 
-@media screen and (min-width: 1175px) {
+@media screen and (min-width: 1024px) {
   .doc > h1.page:first-child {
     margin-top: 2.5rem;
   }
 }
 
-@media screen and (max-width: 1174px) {
+@media screen and (max-width: 1023px) {
   .doc > h1.page:first-child {
     margin-top: 5rem;
   }
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -251,7 +251,7 @@ section.frontpage h2 {
   }
 }
 
-@media screen and (max-width: 1174px) {
+@media screen and (max-width: 1023px) {
   header.frontpage {
     margin-top: var(--navbar-height);
   }
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -251,13 +251,18 @@ section.frontpage h2 {
   }
 }
 
-@media screen and (max-width: 1023px) {
+@media screen and (max-width: 1277px) {
+  header.frontpage {
+    margin-top: var(--navbar-height);
+  }
+
   header.frontpage h1 {
     font-size: 9vw;
   }
 
   header.frontpage svg {
     height: 10rem;
+    top: 8rem;
   }
 
   .split {
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -252,10 +252,6 @@ section.frontpage h2 {
 }
 
 @media screen and (max-width: 1023px) {
-  header.frontpage {
-    margin-top: var(--navbar-height);
-  }
-
   header.frontpage h1 {
     font-size: 9vw;
   }
--- a/antora-ui-camel/src/css/frontpage.css
+++ b/antora-ui-camel/src/css/frontpage.css
@@ -251,7 +251,7 @@ section.frontpage h2 {
   }
 }
 
-@media screen and (max-width: 1277px) {
+@media screen and (max-width: 1174px) {
   header.frontpage {
     margin-top: var(--navbar-height);
   }
--- a/antora-ui-camel/src/css/header.css
+++ b/antora-ui-camel/src/css/header.css
@@ -28,7 +28,7 @@ body {
   padding: 0 0.375rem;
 }
 
-@media screen and (min-width: 1024px) {
+@media screen and (min-width: 1278px) {
   .navbar-end > .navbar-item,
   .navbar-end .navbar-link {
     color: var(--navbar-font-color);
@@ -65,12 +65,13 @@ body {
   border: none;
   outline: none;
   line-height: 1;
-  height: var(--navbar-height);
+  height: 3rem;
   position: relative;
   /* width: var(--navbar-height); */
-  width: 2.75rem;
-  margin-left: auto;
+  width: 2rem;
+  top: 0.5rem;
   padding: 0;
+  order: 8;
 }
 
 .navbar-burger:focus {
@@ -85,15 +86,15 @@ body {
   margin-left: -7px;
   position: absolute;
   top: 60%;
-  width: 15px;
+  width: 25px;
 }
 
 .navbar-burger span:nth-child(1) {
-  margin-top: -6px;
+  margin-top: -10px;
 }
 
 .navbar-burger span:nth-child(2) {
-  margin-top: -1px;
+  margin-top: -3px;
 }
 
 .navbar-burger span:nth-child(3) {
@@ -102,7 +103,7 @@ body {
 
 .navbar-burger.is-active span:nth-child(1) {
   margin-left: -5px;
-  transform: rotate(45deg);
+  transform: rotate(40deg);
   transform-origin: left top;
 }
 
@@ -112,12 +113,13 @@ body {
 
 .navbar-burger.is-active span:nth-child(3) {
   margin-left: -5px;
-  transform: rotate(-45deg);
+  transform: rotate(-40deg);
   transform-origin: left bottom;
 }
 
 .navbar-menu {
   display: none;
+  order: 2;
 }
 
 .navbar-item,
@@ -156,7 +158,7 @@ body {
   margin: 0.25rem 0;
 }
 
-@media screen and (max-width: 1023px) and (min-width: 480px) {
+@media screen and (max-width: 1277px) and (min-width: 480px) {
   .navbar-menu.is-active {
     display: block;
     position: absolute;
@@ -225,14 +227,14 @@ body {
   }
 }
 
-@media screen and (max-width: 479px) {
+@media screen and (max-width: 1277px) {
   .navbar-menu.is-active {
     display: block;
     position: absolute;
     height: 100vh;
     box-shadow: 0 16px 16px 0 rgba(10, 10, 10, 0.1);
     overflow-y: auto;
-    top: 4rem;
+    top: 8rem;
     left: 0;
     right: 0;
     padding: 1rem 2rem;
@@ -295,7 +297,7 @@ body {
   }
 }
 
-@media screen and (min-width: 1024px) {
+@media screen and (min-width: 1278px) {
   .navbar,
   .navbar-menu,
   .navbar-end {
@@ -387,11 +389,19 @@ body {
 
 .navbar-fill {
   flex-grow: 1;
+  order: 3;
+}
+
+.break-row {
+  display: none;
+  flex-basis: 100%;
+  height: 0;
 }
 
 .navbar-search {
   padding: 0.5rem;
   text-align: right;
+  order: 4;
 }
 
 .navbar-search input {
@@ -486,6 +496,7 @@ body {
   line-height: 1.6;
   padding: 0.5rem;
   font-weight: bold;
+  order: 5;
 }
 
 .navbar-tools .brand-icon {
@@ -495,7 +506,7 @@ body {
   fill: var(--navbar-font-color);
 }
 
-@media screen and (max-width: 1023px) {
+@media screen and (max-width: 1277px) {
   .navbar-tools {
     display: flex;
     align-items: flex-end;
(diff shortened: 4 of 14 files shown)


aashnajena:
Hey, how are the search results appearing for <1275px? In the preview, I can see search results appearing only for >1275px screens

antora-ui-camel/src/css/header.css
```diff
@@ -406,6 +423,28 @@ body {
...
+
+@media screen and (max-width: 425px) {
+  .navbar-small input {
+    width: 80vw;
+  }
+}
+
 .navbar-search input:focus-within {
   background-color: #fff;
   border-color: #eaeaec;
```
zregvart, line 450: (not part of this change), we should have used variables here

antora-ui-camel/src/css/header.css
```diff
@@ -426,9 +465,9 @@ body {
   cursor: pointer;
 }
 
-#search_results {
+.search-results {
   top: 3.38rem;
-  background: #fffe;
+  background: #fff;
```
zregvart, line 470: Use variables

antora-ui-camel/src/css/pagination.css
```diff
@@ -29,11 +29,11 @@ nav.pagination span::before {
 }
 
 nav.pagination .prev::before {
```
zregvart, line 31: Style changes in this file should be on a separate commit or in a new pull request.

antora-ui-camel/src/css/site.css
```diff
@@ -22,4 +22,4 @@
 @import 'release.css';
 @import 'misc.css';
 @import 'community.css';
-@import 'projects.css'
+@import 'projects.css';
```
zregvart, line 25: Style changes in this file should be on a separate commit or in a new pull request.

antora-ui-camel/src/css/toc.css
```diff
@@ -51,11 +51,11 @@
   margin: 0;
 }
 
-.toc .toc-menu li[data-level="2"] a {
+.toc .toc-menu li[data-level='2'] a {
   padding-left: 1.25rem;
 }
 
-.toc .toc-menu li[data-level="3"] a {
+.toc .toc-menu li[data-level='3'] a {
```
zregvart, line 58: Style changes in this file should be on a separate commit or in a new pull request.

antora-ui-camel/src/css/vars.css
```diff
@@ -131,11 +131,14 @@
   --footer-height: 23rem; /* empirically corresponding current design of the foooter */
   /* dimensions */
   --navbar-height: calc(73 / var(--rem-base) * 1rem);
+  --toolbar-nav-height: calc(139.5 / var(--rem-base) * 1rem);
+  --panel-height: calc(54 / var(--rem-base) * 1rem);
```
zregvart, line 135: These are some odd values, how did we came to them?

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -6,10 +6,13 @@
   window.addEventListener('load', () => {
     const client = algoliasearch('BH4D9OD16A', '16e3a9155a136e4962dc4c206f8278bd')
     const index = client.initIndex('apache_camel')
-    const search = document.querySelector('#search')
+    const condition = window.innerWidth < 1275
+    const search = condition ? document.querySelector('.search-small') : document.querySelector('.search')
```
zregvart, line 10: I'd prefer if we had a CSS only solution for this, not sure if there's need for two inputs for search. This way we make the implementation much more complex.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -25,17 +28,23 @@
...
     }
 
+    search.addEventListener('click', function (e) {
+      e.stopPropagation()
+    })
+
     search.addEventListener(
       'keyup',
       debounce((key) => {
+        if (condition) document.querySelector('.navbar-burger').style.marginTop = '1rem'
```
zregvart, line 38: I'm not sure that the content security policy will allow changing the style via JavaScript, anyhow we need to minimize the JavaScript and focus on style changes via CSS only.

antora-ui-camel/src/partials/header-content.hbs
```diff
@@ -5,33 +5,46 @@
       <div id="topbar-nav" class="navbar-menu">
         <div class="navbar-end">
           {{#withMenuData}}
-            {{#each @items}}
```
zregvart, line 8: Not sure about this, the code was indented to make it more readable.

antora-ui-camel/src/css/vars.css
```diff
@@ -131,11 +131,14 @@
...
   --toolbar-height: calc(45 / var(--rem-base) * 1rem);
   --drawer-height: var(--toolbar-height);
   --body-top: calc(var(--navbar-height) + 1rem);
   --body-min-height: calc(100vh - var(--navbar-height));
   --nav-height: calc(var(--body-min-height) - var(--toolbar-height));
+  --nav-menu-panel-height: calc(100vh - var(--panel-height));
   --nav-height--desktop: var(--body-min-height);
   --nav-panel-height: calc(var(--nav-height) - var(--drawer-height));
```
zregvart, line 143: This value needs to be adjusted, the height of the search bar needs to be subtracted from the height to fit the explore drawer

antora-ui-camel/src/css/doc.css
```diff
@@ -37,6 +37,12 @@
   margin: 1.5rem 0;
 }
 
+@media screen and (max-width: 1274px) {
+  .static > h1:first-child {
+    margin: 4.05556rem 0 1.5rem 0;
```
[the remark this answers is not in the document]
AemieJ: Alright, I believe those values are already calculated within the `vars.css`, I will directly make use of them.

antora-ui-camel/src/css/doc.css
```diff
@@ -532,6 +538,19 @@
...
+.doc aside {
+  float: right;
+  margin-left: 1rem;
+  width: 25rem;
+  margin-top: 3.5rem;
+}
+
+@media screen and (max-width: 1023px) {
+  .doc.side aside {
+    margin-left: -1rem;
```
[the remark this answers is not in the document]
AemieJ: This is for the category issue that I raised. The box has been shifted towards the right, hence this just shifts it back to original position to the left.

antora-ui-camel/src/css/header.css
```diff
@@ -394,6 +399,18 @@ body {
   text-align: right;
 }
 
+.search-bar {
+  top: 4.1rem;
+  position: sticky;
+  position: -webkit-sticky;
```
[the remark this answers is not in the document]
AemieJ: Yes, we need this 'cause chrome tends to behave in a weird manner at times compared to the other browsers such as firefox and safari. Hence, both the position is required.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
AemieJ: I don't know why it seems complex 'cause the ternary operator does make the code readable. And the reason that two searches are included is just for breakpoint differentiation is 'cause if we were to include the single search input that is present within the nav, it will be more complex to position it the way we want in the design ( I will give it a try though ). Also, in my opinion, the use of two different div for the search design for two breakpoints seems much readable and it's easier to differentiate them using CSS as well.

antora-ui-camel/src/css/blog.css
```diff
@@ -25,6 +25,16 @@ article.blog:first-child {
   margin-top: 3rem;
 }
 
+@media screen and (max-width: 1023px) {
+  .blog.list aside {
+    margin-top: 2.5rem;
+  }
+
+  article.blog-post:first-child {
```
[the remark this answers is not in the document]
AemieJ: No, this change corresponds particularly to the blog post only.

antora-ui-camel/src/css/header.css
```diff
@@ -394,6 +399,18 @@ body {
...
+  z-index: 4;
+}
+
+.navbar-small {
+  display: none;
+  background: #fff;
```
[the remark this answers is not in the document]
AemieJ: It is being displayed for width < 1275px.

antora-ui-camel/src/css/vars.css
AemieJ: Alright, the toolbar nav height is introduced for the width < 1275 as we introduce the whole search design below the actual navbar including the logo and social icons. 

Now with the changes, I observed that within the menu panel height under the docs section, the panel items overflow to the footer due to the height defined. Hence, the panel height is used to prevent the overflow.

AemieJ: Alright, the toolbar nav height is introduced for the width < 1275 as we introduce the whole search design below the actual navbar including the logo and social icons. 

Now with the changes, I observed that within the menu panel height under the docs section, the panel items overflow to the footer due to the height defined. Hence, the panel height is used to prevent the overflow.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
AemieJ: Yes, I experimented and it is possible through HTML and CSS only. I will push the changes for the same.

antora-ui-camel/src/css/header.css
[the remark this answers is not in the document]
zregvart: Should be the same as `sticky`, it can be used unprefixed according to [caniuse.com](https://caniuse.com/#feat=css-sticky) -- 91.35% browsers don't require the prefix.

zregvart:
@AemieJ I've created [this JSFiddle](https://jsfiddle.net/ry2Lhwtd/) to illustrate how this can be done with CSS only. The key is that in flex layout you can specify order of the elements, and that a break can be forced after an element pushing the search div to second row.

AemieJ:
> @AemieJ I've created [this JSFiddle](https://jsfiddle.net/ry2Lhwtd/) to illustrate how this can be done with CSS only. The key is that in flex layout you can specify the order of the elements and that a break can be forced after an element pushing the search div to the second row.

I tried using the method you described in jsfiddle. However, in the output, I observed that at a point it will only push down when there is no more space to be accommodated in the main navbar. I observed at a point where only the navbar is pushed down and the search bar is after the social icon due to ordering, I don't think we want that kind of behavior.

behavior-flex

AemieJ:
@zregvart Another method if we don't want to make use of two search divs, we could increase the height of the navbar and use the translate property of CSS to control it. What do you think about that?

zregvart:
@AemieJ I don't think CSS transform should be used for responsive design. I think responsive design should be based on flexbox. Can you research the example I provided and see if it can be applied to the website?

AemieJ:
@zregvart  I found a method using flex-box on [this jsfiddle](https://jsfiddle.net/kh0n9yge/), please check if you find any loophole. Else, I will move forward on implementing this on the site.

AemieJ:
@zregvart I worked with the flexbox method and applied it to the website, it works out neatly. Please take a look at it and do let me know if anything further is required to be done.

antora-ui-camel/src/css/nav.css
```diff
@@ -118,6 +152,12 @@ html.is-clipped--nav {
   overflow-y: auto;
 }
 
+@media screen and (max-width: 1277px) {
+  .nav-menu {
+    top: 3.5rem;
```
zregvart, line 157: Perhaps derive a value using calc or use a variable here.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -6,10 +6,12 @@
...
     const index = client.initIndex('apache_camel')
-    const search = document.querySelector('#search')
+    const search = document.querySelector('.search')
     const container = search.parentNode
-    const results = document.querySelector('#search_results')
-    const cancel = document.querySelector('#search-cancel')
+    const results = document.querySelector('.search-results')
+    const cancel = document.querySelector('.search-cancel')
+    const className = 'navbar-search'
+    const hiddenClass = 'navbar-search results-hidden'
```
zregvart, line 14: Do we still need these changes to the JS code now that we have only one search input?

antora-ui-camel/src/partials/header-content.hbs
```diff
@@ -6,32 +6,39 @@
...
           {{/withMenuData}}
         </div>
       </div>
       <div class="navbar-fill"></div>
+      <div class="break-row"></div>
       <div class="navbar-search results-hidden">
         <input id="search" class="search" placeholder="Search" autocomplete="off">
-        <img src= "/_/img/cancel.svg" alt="Clear" id="search-cancel">
-        <div id="search_results"></div>
+        <img src= "/_/img/cancel.svg" alt="Clear" id="search-cancel" class="search-cancel">
```
zregvart, line 29: Do we need the `search-cancel` class here?

antora-ui-camel/src/partials/header-content.hbs
```diff
@@ -6,32 +6,39 @@
...
+        <div id="search_results" class="search-results"></div>
```
zregvart, line 30: Do we need the `search-results` class here?

antora-ui-camel/src/partials/nav.hbs
```diff
@@ -1,5 +1,9 @@
 <div class="nav-container"{{#if page.component}} data-component="{{page.component.name}}" data-version="{{page.version}}"{{/if}}>
+  {{#if (eq page.component.name 'components')}}
+  <aside class="nav nav-components">
+  {{else}}
   <aside class="nav">
+  {{/if}}
```
zregvart, line 6: I think this could be simpler, all `aside.nav` should cater to the same positioning, not sure why components would be different, perhaps the parent of this aside (`.nav-container`) should be positioned instead.

antora-ui-camel/src/css/header.css
```diff
@@ -387,11 +407,19 @@ body {
 
 .navbar-fill {
   flex-grow: 1;
+  order: 3;
+}
+
+.break-row {
+  display: none;
+  flex-basis: 100%;
+  height: 0;
```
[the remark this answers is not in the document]
AemieJ: The height: 0 is the reason the row will break. The display: none is only present as we don't need this change above the breakpoint.

antora-ui-camel/src/css/header.css
```diff
@@ -417,18 +445,18 @@ body {
   }
 }
 
-#search-cancel {
+.search-cancel {
```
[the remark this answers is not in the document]
AemieJ: @zregvart yes, I will change it for both search cancel and search results to use it by Id directly.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
AemieJ: No, I will change to use search cancel and search results by Id and revert the changes for this to normal.

AemieJ:
@zregvart I will take care of it in the future PR to only include one change.


Other than that, as I have read your comments I have to make use of variables and revert the algolia search bundle to normal as well as remove class names for search cancel and search results.

 I don't realize what part is complex yet as I now use only one input and make use of flex box and everything else is css based on the breakpoint?

antora-ui-camel/src/css/header.css
```diff
@@ -28,7 +34,7 @@ body {
   padding: 0 0.375rem;
 }
 
-@media screen and (min-width: 1024px) {
+@media screen and (min-width: 1278px) {
```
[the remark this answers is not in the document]
AemieJ: Alright, yes as the navbar menu will be changed, the breakpoint will change accordingly.

zregvart:
> I don't realize what part is complex yet as I now use only one input and make use of flex box and everything else is css based on the breakpoint?

The complexity I see is from adding conditionals to the navigation panel rendering the components reference and that a lot of positioning seems to be done by finding the values and putting them in, rather than deriving them from variables/using `calc`. What I see as a problem is that if we change, say the height of the navbar, this positioning needs to be recalculated again. There's also the element that was added to ensure the break on smaller screens, I'm not sure that it's needed. Using flex layout with `float`. Overall if you look at the changes here and the outcome from those it seems like a lot of changes for not that many changes in the output. I think if we were to redo this knowing the principles that are needed for this change, carefully adding/changing only the CSS that needs to be changed, we'd end up with 33%-50% of the CSS we have now.

Anyhow don't let that discourage you, this is excellent work, keep polishing this, we can use a lot of it and hopefully merge it soon.

antora-ui-camel/src/css/header.css
```diff
@@ -59,18 +65,26 @@ body {
   height: var(--navbar-height);
 }
 
+@media screen and (max-width: 1277px) {
+  .navbar-brand {
+    flex-wrap: wrap;
```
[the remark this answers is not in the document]
AemieJ: I will look into this as I didn't face this issue.

AemieJ:
@zregvart okay as I know what needs to be done with flex box I will try to do this from scratch. It might take a day or 2.

AemieJ:
@zregvart I have done highly specific changes to it from scratch solving each part. Although the breakpoint is defined right now is at 1278px but let's wait for #397 to get merged, then I could change the breakpoint for it. 

It works fine as I have checked. Please check for any loopholes if you can find.

AemieJ:
@zregvart With the current changes to the navbar menu, the breakpoint changes from 1278px to 1175px.

zregvart:
@AemieJ what if we shrunk the logo width in the nav bar to 4rem and reduced the padding between menu items from 2rem to 1.5rem, then at 1024px we can still have the _big_ menu:

Screenshot_2020-06-29 Home - Apache Camel

AemieJ:
@zregvart that's a good suggestion if we just want the logo to appear in the nav menu. It will space as well.

Delawen:
That's good news!

aashnajena:
Looks good!

antora-ui-camel/src/css/header.css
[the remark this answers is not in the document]
zregvart: I'm not sure that elements that have `display:none` have a hight, do they?

antora-ui-camel/src/css/blog.css
```diff
@@ -223,3 +223,13 @@ article.blog p {
 .blog .page-item.active a {
   color: var(--color-white);
 }
+
+@media screen and (max-width: 1023px) {
+  article.blog:first-child {
```
zregvart, line 228: This selector might be a bit problematic as it might target anything (including an element that's absolutely positioned). I'm fairly certain that to accommodate the second row on the top level navigation we need to move or add the padding/margin globaly. I think this change needs to go to `base.css` and cover all pages. We should also base this from the `--navbar-height` variable like we do for the `--body-top`. Remember these variables can change and then `5rem` is not accurate.

antora-ui-camel/src/css/blog.css
```diff
@@ -223,3 +223,13 @@ article.blog p {
...
+    margin-top: 5rem;
+  }
+
+  .blog.list aside {
+    margin-top: 4.5rem;
```
zregvart, line 233: If we move the whole `--body-top` then we don't need to reposition the individual elements, like the category menu here.

antora-ui-camel/src/css/doc.css
```diff
@@ -37,12 +37,18 @@
...
 
-@media screen and (min-width: 769px) {
+@media screen and (min-width: 1024px) {
   .doc > h1.page:first-child {
     margin-top: 2.5rem;
   }
 }
 
+@media screen and (max-width: 1023px) {
+  .doc > h1.page:first-child {
```
zregvart, line 47: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -252,12 +252,17 @@ section.frontpage h2 {
 }
 
 @media screen and (max-width: 1023px) {
+  header.frontpage {
+    margin-top: var(--navbar-height);
```
zregvart, line 256: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -252,12 +252,17 @@ section.frontpage h2 {
...
+  }
+
   header.frontpage h1 {
     font-size: 9vw;
   }
 
   header.frontpage svg {
     height: 10rem;
+    top: 8rem;
```
zregvart, line 265: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/css/header.css
```diff
@@ -59,18 +65,26 @@ body {
...
+}
+
 .navbar-burger {
   color: var(--navbar-font-color);
-  background: none;
+  background: var(--navbar-background);
   border: none;
   outline: none;
   line-height: 1;
-  height: var(--navbar-height);
```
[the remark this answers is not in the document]
zregvart: Use the `--navbar-height` variable instead.

antora-ui-camel/src/css/header.css
```diff
@@ -225,14 +227,14 @@ body {
...
-@media screen and (max-width: 479px) {
+@media screen and (max-width: 1023px) {
   .navbar-menu.is-active {
     display: block;
     position: absolute;
     height: 100vh;
     box-shadow: 0 16px 16px 0 rgba(10, 10, 10, 0.1);
     overflow-y: auto;
-    top: 4rem;
+    top: 8rem;
```
zregvart, line 237: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/css/header.css
```diff
@@ -318,6 +320,12 @@ body {
     padding: 1rem 2rem 1rem 0;
   }
 
+  @media screen and (max-width: 1080px) {
+    .navbar-item {
+      padding-right: 1.5rem;
+    }
+  }
```
zregvart, line 327: I'd change this in higher widths, no need for a lot of padding, 1.5rem should look okay in all resolutions.

antora-ui-camel/src/css/header.css
```diff
@@ -351,6 +359,12 @@ body {
     margin-left: 2rem;
   }
 
+  @media screen and (min-width: 1024px) and (max-width: 1080px) {
+    .navbar-end {
+      margin-left: 1rem;
```
zregvart, line 364: This is to change the left margin between the top level navigation (i.e. "Blog") and the logo? I think we can make that for all resolutions.

antora-ui-camel/src/css/static.css
```diff
@@ -1,6 +1,5 @@
 .static {
   margin: var(--static-margin);
-  max-width: var(--static-max-width);
```
zregvart, line 3: Not sure about this change, why are we making it?

antora-ui-camel/src/css/static.css
```diff
@@ -9,6 +8,12 @@
   }
 }
 
+@media screen and (max-width: 1023px) {
+  .static {
+    margin-top: 5rem;
```
zregvart, line 13: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/css/toc.css
```diff
@@ -8,6 +8,13 @@
   top: var(--toc-top);
 }
 
+@media screen and (min-width: 1024px) and (max-width: 1174px) {
+  .toc.sidebar .toc-menu {
+    margin-top: 4rem;
```
zregvart, line 13: Could be unnecessary if we move the `--body-top`.

antora-ui-camel/src/partials/header-content.hbs
```diff
@@ -23,6 +23,7 @@
         </div>
       </div>
       <div class="navbar-fill"></div>
+      <div class="break-row"></div>
```
zregvart, line 26: Doesn't seem to make a difference, I've removed this from the DOM in developer tools and resized from 1023px to 300px, and the layout remained the same.

AemieJ:
@zregvart  I simplified a bit further using your suggestion. however certain changes are necessary.

zregvart:
Awesome @AemieJ, great effort, and thank you for tolerating my nitpickery!

AemieJ:
@zregvart well it certainly helped with simplifying the code as much as possible, so I am glad and thanks!

PR #400, opened 2020-06-18
CAMEL-15202: fix blog post pagination design

zregvart:
This reintroduces the design we had before rebasing on the latest Antora
Default UI (#383). Now the changes are affecting only the blog section
and further rebases should not interfere with the pagination design.


zregvart:
Preview looks good, so I merged.

PR #401, opened 2020-06-19
CAMEL-15202: Add the custom scrollbar for nav menu panel

AemieJ:
This introduces the scrollbar design for the nav menu panel under the documentation section after the rebasing in PR #383 altered the change in design. I change the contrast color with respect to the basic contrast ratio for scrollbars.

### DIFF
--- a/antora-ui-camel/src/css/nav.css
+++ b/antora-ui-camel/src/css/nav.css
@@ -137,6 +137,13 @@ html.is-clipped--nav {
   background: rgb(136, 136, 136);
 }
 
+@media screen and (max-width: 1023px) {
+  .nav-menu::-webkit-scrollbar {
+    width: 0;
+    background: transparent;
+  }
+}
+
 .nav-menu h3.title {
   color: var(--nav-heading-font-color);
   font-size: inherit;
--- a/antora-ui-camel/src/css/nav.css
+++ b/antora-ui-camel/src/css/nav.css
@@ -123,18 +123,18 @@ html.is-clipped--nav {
 }
 
 .nav-menu::-webkit-scrollbar-track {
-  background: var(--nav-scrollbar-color);
+  background: var(--scrollbar-track-color);
   border-radius: 3.125rem;
 }
 
 .nav-menu::-webkit-scrollbar-thumb {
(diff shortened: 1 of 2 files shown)


AemieJ:
@zregvart  I did changes similar to the one in PR #287 that focused on the custom scrollbar as well.

antora-ui-camel/src/css/vars.css
```diff
@@ -54,6 +54,8 @@
   /* nav */
   --nav-background: var(--panel-background);
   --nav-border-color: var(--color-smoke-50);
+  --nav-scrollbar-color: var(--color-gray-10);
+  --nav-scrollbar-active-color: var(--color-gray-50);
```
zregvart, line 58: Can we make this on same lines and with the same name as in the [Default Antora UI](https://gitlab.com/antora/antora-ui-default/-/blob/master/src/css/vars.css), this would make it easier to rebase.

antora-ui-camel/src/css/nav.css
```diff
@@ -118,6 +118,32 @@ html.is-clipped--nav {
...
+  background: var(--nav-scrollbar-active-color);
+}
+
+@media screen and (max-width: 1023px) {
+  .nav-menu::-webkit-scrollbar {
+    width: 0;
+    background: transparent;
+  }
+}
+
```
zregvart, line 146: If we could make this changes align with [the Default Antora UI](https://gitlab.com/antora/antora-ui-default/-/blob/master/src/css/nav.css#L91-105), that would be awesome, much easier to rebase then.

antora-ui-camel/src/css/vars.css
AemieJ: Yes, I used it similar to it and included additional ones as well one for track and one for when the thumb of the scrollbar is active.

antora-ui-camel/src/css/nav.css
AemieJ: The thing is that in the default antora UI, they make use of the `nav-panel-menu` however in our case, we make use of the `nav`  'cause under the component section, we need to avoid the scrollbar for the `quick lookup` as it is included within `nav-panel-menu`.

PR #405, opened 2020-06-24
update irc-room.md

AemieJ:


davsclaus:
Oh this is some old content - we do not use irc anymore. Maybe we can delete this doc file, and remove it from any index.

AemieJ:
@davsclaus I will do that then if it isn't required anymore.

davsclaus:
Thanks yeah please go ahead and remove it

AemieJ:
And also, didn't find any `index.md` with the context of irc itself.

PR #413, opened 2020-06-25
Automate the process of building the website using Makefile

AemieJ:
To ease the building of the website and automate the process with a single step to build and preview, I believe the use of Makefile is a good solution. In my opinion, any new contributor or current contributor can easily build the projects by using the `make` command by following the instructions I updated for README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,23 +3,20 @@ all: build preview
 
 install: package.json
 		@echo "Installation started." && \
-		cd antora-ui-camel && \
-		yarn install && cd .. && \
+		yarn --cwd antora-ui-camel/ install && \
 		yarn install
 
 format: package.json
 		@echo "Formatting has started" && \
-		cd antora-ui-camel && \
-		yarn format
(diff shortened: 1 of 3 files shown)


djencks:
Can't you do this just as well with a gulp file without introducing the use of yet another tool?

AemieJ:
@djencks I will look into that. 👍

AemieJ:
@djencks Yes, there is a library within gulp( **child_process** ) with which it would be possible however for this case ( just building of the website easily ), I thought Makefile would be a better option 'cause it is really easy to maintain as well.

zregvart:
Thanks for understanding @AemieJ, I think simplifying the build is a great idea, and your proposal made me think that we can use Yarn workspaces to make it a bit easier. See https://issues.apache.org/jira/browse/CAMEL-15249 for the idea.

AemieJ:
@zregvart I will give a look at it. It seems a more suitable option for this case.

PR #415, opened 2020-06-26
fix: provide padding between text field and the cancel button

AemieJ:
* Previously, when long text is written within the search bar input it hides behind the cancel button.

searchbar-prev

* Fixed it by added sufficient padding to avoid this issue.

searchbar-next


aashnajena:
We should also have the cancel button at the vertical center of the input box. It's visibly lower than that.

AemieJ:
@aashnajena That will be taken care of under the search design implementation.

PR #416, opened 2020-06-28
refactor: cleanup frontpage.css

aashnajena:
Cleaned up frontpage.css to group css rules together and scope them better. Also made some changes to ensure content is better spaced, headings/margins are consistent for all frontpage sections and boxes from "projects" and "functionalities" section are aligned.

The icons in the "functionalities" section are used as images and use a complicated "filter" property to colour them blue. On my laptop, that colour doesn't really match the camel blue. I tried to use them as `<svg> <use href="path" /> </svg>` but this isn't working for me.

### DIFF
--- a/antora-ui-camel/src/css/docs.css
+++ b/antora-ui-camel/src/css/docs.css
@@ -1,31 +1,37 @@
-.docs h2 {
+.docs {
+  padding: 1rem 3rem;
+}
+
+.project h2 {
   text-align: center;
+  padding-bottom: 0;
 }
 
 .camel-project {
-  width: calc(100% - 2.2rem);
+  width: calc(100% - 3rem);
   display: inline-flex;
   border: 1px solid var(--color-smoke-90);
   flex-direction: column;
   align-items: center;
   text-align: center;
   padding: 1rem 1rem 2rem 1rem;
-  margin: 1rem;
+  margin: 1.5rem;
 }
 
 .camel-project .project {
   border: none;
+  padding: 0;
 }
 
 .project {
   display: inline-flex;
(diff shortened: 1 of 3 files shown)


AemieJ:
The functionality section was designed in a manner to occupy lesser space however with the current changes, it is just a box design covering lots of space as it did before. Except for this, it is good.

aashnajena:
@AemieJ can you specify where a lot of space is getting occupied? I attempted to align the sections correctly, I didn't want to change the design.

aashnajena:
@zregvart do we need to fix the "functionalities" section icons?

AemieJ:
https://issues.apache.org/jira/browse/CAMEL-15161?jql=project%20%3D%20CAMEL%20AND%20component%20%3D%20website 

This issue dealt with reducing the space from the old functionality to new functionality as you can observe in the attachments. However, with your changes, if I just remove the border from the box it's the same issue as before. 

The reason for this issue was to allow the user to see functionality in one look when it comes across the title but with yours, it reverts to same padding as before and covers too much space.

zregvart:
> @zregvart do we need to fix the "functionalities" section icons?

We can, in a separate PR or in this one. What we do instead for the icons in the top navigation is to change the image background to change the fill color, we also use sprites to reduce the amount of requests to the server. See `brand-logos.svg`.

The [CAMEL-15161](https://issues.apache.org/jira/browse/CAMEL-15161) was about seeing that there is more content (like the Functionality section) above the fold on initial page load. From what I can see we can't for resolutions with height less than or equal to 768px. Not sure what above the fold height is the minimum we need to support.

zregvart:
Thanks!

PR #418, opened 2020-06-29
fix: nav panel explore toggling on click

AemieJ:
* The nav explore menu doesn't trigger on click. So, I resolved with a quick fix. Yet I wanted to an opinion that if the nav-explore-menu is clicked and the menu appears, do you want to disable clicking of any nav-menu items on the panel or not?

### DIFF
--- a/antora-ui-camel/src/js/01-nav.js
+++ b/antora-ui-camel/src/js/01-nav.js
@@ -39,8 +39,10 @@
   })
 
   nav.querySelector('.nav-panel-explore').addEventListener('click', function () {
-    var currentPanel = nav.querySelector('.nav-panel-explore')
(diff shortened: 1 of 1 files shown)


antora-ui-camel/src/js/01-nav.js
```diff
@@ -38,11 +38,9 @@
     }
   })
 
-  nav.querySelector('.context').addEventListener('click', function () {
-    var currentPanel = nav.querySelector('.is-active[data-panel]')
```
zregvart, line 42: Perhaps the only change that's needed is to change this line to:
```javascript
var currentPanel = nav.querySelector('.is-active + div[data-panel]')
```

zregvart:
Thanks!

PR #419, opened 2020-06-30
fix: adjustments for the category box within blog section

AemieJ:
### ISSUE OBSERVED
old-category

### DIFF
--- a/antora-ui-camel/src/css/blog.css
+++ b/antora-ui-camel/src/css/blog.css
@@ -14,6 +14,12 @@
   line-height: 2rem;
 }
 
+@media screen and (max-width: 1023px) {
+  .blog.list aside {
+    margin-top: 0;
+  }
+}
+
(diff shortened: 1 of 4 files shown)


zregvart:
@AemieJ can you describe the problem with more detail, I'm not sure what is addressed here. The screenshot seems odd I can't see the "Edit this page"/"Back to top" links if I reduce the width below 1024px in Firefox or in Chromium.

AemieJ:
@zregvart The issue is that within the blogs section when the user scrolls to the bottom and clicks on the toolbar toggle, the menu appears with all the categories, however, there is a margin-top and it doesn't cover the full page and appears in the manner as the screenshot, it is very odd behavior and not ideal. So with the changes, I make it so that the whole category box covers the full width and there is no margin-top issue.

zregvart:
All clear now, thanks @AemieJ!

PR #422, opened 2020-07-04
fix: alignment of the documentation page

AemieJ:
The responsiveness for the documentation page wasn't taken into complete account for the smaller screen width. Hence, this PR resolves it.

### BEFORE
bug-layout-responsive

### AFTER
fix-layout-responsive

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -50,12 +50,19 @@
 }
 
 .doc h2:not(.discrete) {
-  border-bottom: 1px solid var(--section-divider-color);
   margin-top: 1rem;
   margin-left: -1rem;
   margin-right: -1rem;
   padding: 0.4rem 1rem 0.1rem;
   overflow-wrap: break-word;
+}
+
+.doc h2:not(.discrete)::after {
+  content: "";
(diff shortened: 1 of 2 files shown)


zregvart:
The `.yarn/install-state.gz` files should not be part of this commit.

aashnajena:
I think instead of doing this we should remove the extra side padding for smaller screens. If we make the padding under .docs 1% and remove the extra padding for "user manual" and "components" under Camel Core for smaller screens, we should be fine.

AemieJ:
> I think instead of doing this we should remove the extra side padding for smaller screens. If we make the padding under .docs 1% and remove the extra padding for "user manual" and "components" under Camel Core for smaller screens, we should be fine.

Taking your opinion into account, removing padding does create extra spacing however the border-bottom isn't the best bottom as it tends to go out of the frame and I need to make use of the word-break for preventing the link overflowing out of the frame.

zregvart:
This looks good to me, @aashnajena or @delawen any other issues you see?

aashnajena:
LGTM

Delawen:
Good! :+1:

zregvart:
Thanks!

PR #423, opened 2020-07-06
CAMEL-14952: Better search on the website

aashnajena:
I have linked my own Algolia index to this PR for now, so that everyone can view the changes in search results. 
- Search results are now grouped by higher level headings eg. "Camel Components" or "User Manual" or "Camel Kafka Connector". 
- For Hugo pages, search results are grouped by H1, because that itself is the highest level of heading eg "Articles" or "Books"
- I wanted to make the config file as simple as possible, and depend on the sitemap instead of the URLs. This is what I'm using :

  "sitemap_urls": [
    "https://camel.apache.org/sitemap.xml"
  ],
  "selectors": {
    "lvl0": ".toolbar ul li:nth-child(1) a",
    "lvl1": ".version-menu a.is-current",
    "lvl2": "article h1",
    "lvl3": "article h2",
    "text": "article p, article li"
  }

### DIFF
--- a/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
+++ b/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
@@ -4,8 +4,8 @@
   const algoliasearch = require('algoliasearch/lite')
 
   window.addEventListener('load', () => {
-    const client = algoliasearch('QDOJ4LJQCC', '0fa28c3acc4d5f383b85ae4ad035ac9c')
-    const index = client.initIndex('aashna')
+    const client = algoliasearch('BH4D9OD16A', '16e3a9155a136e4962dc4c206f8278bd')
+    const index = client.initIndex('apache_camel')
     const search = document.querySelector('#search')
     const container = search.parentNode
     const results = document.querySelector('#search_results')
@@ -53,19 +53,15 @@
             const data = hits.reduce((data, hit) => {
               const d = {}
               d.url = hit.url
-              var hierarchy = Object.values(hit.hierarchy).filter((lvl) => lvl !== null)
-              const section = hierarchy[0]
-              var breadcrumbs = hierarchy.slice(1).join(' &raquo; ')
-              if (breadcrumbs !== '') {
-                d.breadcrumbs = breadcrumbs
-              } else {
-                d.breadcrumbs = section
-              }
-              if (hit._snippetResult !== undefined) {
-                d.snippet = hit._snippetResult.content.value
-              } else {
-                d.snippet = ''
-              }
+              var section = hit.hierarchy.lvl1
+              if (hit.hierarchy.lvl0 !== null) section = section + ' [' + hit.hierarchy.lvl0 + ']'
+              var breadcrumbs = Object.values(hit.hierarchy)
+                .slice(2)
+                .filter((lvl) => lvl !== null)
+                .join(' &raquo; ')
+
+              d.breadcrumbs = ((breadcrumbs !== '') ? breadcrumbs : section)
+              d.snippet = hit._snippetResult.content.value + '...'
 
               data[section] = data[section] || []
               data[section].push(d)
--- a/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
+++ b/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
@@ -4,8 +4,8 @@
   const algoliasearch = require('algoliasearch/lite')
 
   window.addEventListener('load', () => {
-    const client = algoliasearch('BH4D9OD16A', '16e3a9155a136e4962dc4c206f8278bd')
-    const index = client.initIndex('apache_camel')
+    const client = algoliasearch('QDOJ4LJQCC', '0fa28c3acc4d5f383b85ae4ad035ac9c')
+    const index = client.initIndex('aashna')
     const search = document.querySelector('#search')
     const container = search.parentNode
     const results = document.querySelector('#search_results')
@@ -53,14 +53,13 @@
             const data = hits.reduce((data, hit) => {
               const d = {}
               d.url = hit.url
-              var breadcrumbs = Object.values(hit.hierarchy)
-                .slice(1)
-                .filter((lvl) => lvl !== null)
-                .join(' &raquo; ')
+              var hierarchy = Object.values(hit.hierarchy).filter((lvl) => lvl !== null)
+              const section = hierarchy[0]
+              var breadcrumbs = hierarchy.slice(1).join(' &raquo; ')
               if (breadcrumbs !== '') {
                 d.breadcrumbs = breadcrumbs
               } else {
-                d.breadcrumbs = hit.hierarchy.lvl0
+                d.breadcrumbs = section
               }
               if (hit._snippetResult !== undefined) {
                 d.snippet = hit._snippetResult.content.value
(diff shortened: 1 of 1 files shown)


antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -4,8 +4,8 @@
   const algoliasearch = require('algoliasearch/lite')
 
   window.addEventListener('load', () => {
-    const client = algoliasearch('BH4D9OD16A', '16e3a9155a136e4962dc4c206f8278bd')
-    const index = client.initIndex('apache_camel')
+    const client = algoliasearch('QDOJ4LJQCC', '0fa28c3acc4d5f383b85ae4ad035ac9c')
+    const index = client.initIndex('aashna')
```
AemieJ, line 8: The name should be `apache_camel`, I believe.

aashnajena: As I mentioned, I've linked my personal index here, which is named "aashna". This is just for showing everyone how the search results will look with the new configuration. These two lines are not supposed to be merged, I will revert them before merging.

aashnajena:
@zregvart please trigger the preview now, PR is ready for review

AemieJ:
I have a few comments on these changes : 
* You have only included the first child of the toolbar list and neglecting the rest. However, if you search for `Bean`, there is Bean under `Components` as well as `EIP`. Hence, I suggested including the links of the toolbar to avoid any cases of such behavior. If we were to use specific selectors, there would be no loophole. 

* Another suggestion: Initially we made changes to the docs pages so that the title wasn't the only thing shown in the search result. I believe we can do the same with blog pages. The hierarchy I would prefer seeing the Blog as - 
``` 
# This being for the blog pages
Blog
-------
Category Name > Title of the blog
```
* Another case that is not taken care of is that when you search a particular blog, it gives results from the `https://camel.apache.org/category/<title>` as well. To pass this case, I was thinking that we can't avoid crawling on categories as it could be the page where the user wants to land on however to avoid confusion on whether it is the blog page or the category page, we could provide specific layout such as - 
```
# This being for the category pages
Category Name
-------
Title of the Blog
```

aashnajena:
About the first point, we can include the toolbar links without specifying selectors, but I don't think it's a good idea to have all the links from the toolbar. That would make the headings very long and repetitive for most components. I'd wait for some more reviews on this part. We need to make a call about how much of the hierarchy we want to show. My intent is to have one high-level topic - which should be used for **grouping** results, and the rest of the headings to consist of h1 and h2 only. 

I agree that we need to treat blogs differently. I'll try to push some changes for that soon!

aashnajena:
@zregvart @Delawen @AemieJ I've made some changes and pushed a new index. This iteration fixes a lot of previous issues and has the following features :

- Pages are grouped by versions. Versions are displayed along with the top-level heading. Versions should appear for all documentation pages now 
- Each search result is grouped under User Manual [version], Camel Components [version], Blog, Books, Downloads, Community, or some sub-project [version].  
- No duplicate results are shown. Try searching for "Components" or "Kafka" or an intermediate hierarchy level like "Architecture". You shouldn't be seeing duplicate results as was the case earlier.
- Each result contains a snippet of text/abstract with relevant words highlighted. Earlier, abstract was not appearing in many cases.
- Intermediate levels of hierarchy are visible in the header/title of each search result. As far as I have checked, there are no missing levels. Earlier I had observed some inconsistencies with pages like "Properties" 

Known problems: For Blog, the < > arrows are showing up in abstract. I have removed this in the final config file, however, it's too lengthy for me to create the whole index again. Please overlook this detail. Please review the changes this week, because I'm again on a free Algolia trial, as our community account does not allow me to push >10k records for test app.

AemieJ:
Include `index.html` within `stop_urls` list within the config.

error

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -4,8 +4,8 @@
   const algoliasearch = require('algoliasearch/lite')
 
   window.addEventListener('load', () => {
-    const client = algoliasearch('BH4D9OD16A', '16e3a9155a136e4962dc4c206f8278bd')
-    const index = client.initIndex('apache_camel')
+    const client = algoliasearch('XW4Y7DJE2Z', '2423e9780121b010cad442bfb45a4714')
+    const index = client.initIndex('camel_test')
```
Delawen, line 8: Do we want to call it "test"?

Delawen:
I love PR that the total count remove lines :)

I like it.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
aashnajena: These two lines of changes will be reverted before merging. "camel_test" is the name of my temporary local index which I have linked here so that everyone can see the improved search results.

zregvart: > Do we want to call it "test"?

This is against the test index configuration so we can see the different index configuration in the preview, before we merge we'll revert to the production values. That reminds me we also need to create a PR with the configuration change at Algolia. @aashnajena, @AemieJ do we need to merge the configuration change at Algolia before we merge this change?

aashnajena: This PR is so that everyone can search for keywords on the [preview](https://deploy-preview-423--camel.netlify.app/) and comment on whether we are getting better results than before (in terms of grouping results, duplicates, versions, etc).

aashnajena: @zregvart Yes, we need to make a PR on Algolia's repository first. If you're satisfied with the current search results, should I go ahead and make that PR?

AemieJ:
Based on `index.html` it creates duplicate results for certain queries as `/` and `/index.html` leads to the same page so if Aashna can check by including `index.html` within `stop_urls`, then I believe it will possibly all the duplicate results.

Example - [Community page with index.html within webpage domain](https://camel.apache.org/manual/latest/index.html#_community) and [same community page without the index.html within webpage domain](https://camel.apache.org/manual/latest/#_community)

@zregvart I also was thinking of a few tweaks in the design and presentation for the search results but I wanted to tell after this search was updated with max possible efficiency.

aashnajena:
@AemieJ yes I have included `index.html` under stop_urls.

aashnajena:
I have opened a [Pull Request](https://github.com/algolia/docsearch-configs/pull/2097) on Algolia's Docsearch Configs repository. @zregvart They require a maintainer of the camel-website repository to approve those changes.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -53,22 +53,16 @@
...
-                d.breadcrumbs = hit.hierarchy.lvl0
-              }
-              if (hit._snippetResult !== undefined) {
-                d.snippet = hit._snippetResult.content.value
-              } else {
-                d.snippet = ''
-              }
-
-              const section = hit.hierarchy.lvl0
+
```
AemieJ, line 62: There is a trailing space here according to netlify, same case with line 65.  Remove it and then we could test the preview.

aashnajena:
@zregvart can we merge this? We need these changes for Search to work correctly on the website.

zregvart:
Thanks!

PR #425, opened 2020-07-08
fix: toc sidebar width explicitly mentioned

AemieJ:
### Issue-1: Width occupied by toc sidebar
toc-issue-1

### Resolve-1: Explicitly providing a min-width for the toc sidebar
toc-resolve-1

### Issue-2: Unnecessary scrollbar present for a toc sidebar menu. (Uglifies the web page and it isn't required either)
toc-issue-2

### Resolve-2: Make the scrollbar transparent (browser-specific) so the scrolling will take place without the need to see a scrollbar

### DIFF
--- a/antora-ui-camel/src/css/main.css
+++ b/antora-ui-camel/src/css/main.css
@@ -21,6 +21,17 @@
 
   aside.toc.sidebar {
     flex: 0 0 var(--toc-width);
+    min-width: 15rem;
+  }
+
+  .toc.sidebar .toc-menu ul {
+    -ms-overflow-style: none; /* Internet Explorer 10+ */
+    scrollbar-width: none; /* Firefox */
+  }
+
+  .toc.sidebar .toc-menu ul::-webkit-scrollbar {
+    width: 0;
+    background: transparent;  /* Chrome/Safari/Webkit */
   }
 }
 


zregvart:
I'm not sure #1 is an issue, how do I reproduce it? On smaller screens the TOC should be on top of the page like here:
Screenshot_2020-07-08 Quartz Apache Camel

I would not introduce browser specific syntax if not absolutely necessary and I don't think hiding the scrollbar is good for accessibility. We have pages ([example](https://camel.apache.org/manual/latest/using-propertyplaceholder.html)) with a lot of content/headings so navigating them might require the scrollbar to be present. In that case the issue I see is that the TOC overlaps with the "back to top"

AemieJ:
@zregvart  Issue 1 is raised at a screen width of 1024px for the Quartz page. That's why I find it important to include a width for toc. Yes, scrollbars could help with navigation but I would prefer seeing a custom scrollbar than the scrollbar that appears for the toc menu. I don't mind the scrollbar to be visible but it would be better if the scrollbar was thin for this purpose and not as the one in the image I showed. 

I have observed the overlapping but I am not sure If it is considered an issue entirely, I even observed it in antora docs as well when I was looking through it.

zregvart:
Thanks!

PR #431, opened 2020-07-19
fix: responsiveness for docs section pages

AemieJ:
The [docs](https://camel.apache.org/docs/) page has comparatively different padding arrangement for the breakpoint to create a responsive and structured layout for the page however as `docs` is the page section used, the same padding affects its corresponding [building](https://camel.apache.org/docs/building) and [sources](https://camel.apache.org/docs/sources) pages below the breakpoint ( 1024px ) and it creates unresponsiveness for both these pages. 

Hence, I made the required changes that keep the structured layout for the docs pages and fix the responsiveness for both the above-mentioned docs-section pages.

### Issue within responsiveness and presentation of the building page below 1024px 
building-docs

### DIFF
--- a/antora-ui-camel/src/css/docs.css
+++ b/antora-ui-camel/src/css/docs.css
@@ -62,27 +62,19 @@
 }
 
 @media screen and (max-width: 1023px) {
-  .docs .project {
-    width: calc(100% - 3.2rem);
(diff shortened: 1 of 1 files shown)


zregvart:
Thanks!

PR #432, opened 2020-07-20
CAMEL 15296: Create sitemap for Camel website

aashnajena:
This is the first draft for adding sitemaps to the website. On the preview, please go to `/sitemap/` to view the sitemaps, since there's no direct link to it present anywhere as of now. Also, I have linked my fork of the Camel repo, so that you view sitemaps for user manual and components. We will have to add a sitemap.adoc file in each of the sub-projects as well. sitemap.adoc in each repo would just have the title and page-layout attribute, no content.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -6,7 +6,7 @@ site:
 content:
   sources:
 
-    - url: git@github.com:apache/camel.git
+    - url: git@github.com:aashnajena/camel.git
       branches: master
       start_paths:
         # manual
@@ -19,7 +19,7 @@ content:
         # main components doc
         - docs/components
         
-    - url: git@github.com:apache/camel.git
+    - url: git@github.com:aashnajena/camel.git
       branches: camel-3.4.x
       start_paths:
         # eip
@@ -30,7 +30,7 @@ content:
         # main components doc
         - docs/components
 
-    - url: git@github.com:apache/camel.git
+    - url: git@github.com:aashnajena/camel.git
       branches: camel-2.x
       start_paths:
         # main components doc
--- a/antora-ui-camel/src/css/site.css
+++ b/antora-ui-camel/src/css/site.css
@@ -8,7 +8,7 @@
 @import 'toolbar.css';
 @import 'breadcrumbs.css';
 @import 'page-versions.css';
-@import "toc.css";
+@import 'toc.css';
 @import 'doc.css';
 @import 'static.css';
 @import 'header.css';
@@ -23,3 +23,4 @@
 @import 'misc.css';
 @import 'community.css';
 @import 'docs.css';
+@import 'sitemap.css'
--- a/antora-ui-camel/src/css/sitemap.css
+++ b/antora-ui-camel/src/css/sitemap.css
@@ -1,0 +1,16 @@
+.sitemap > ul > li > ul,
+ul.sitemap-list {
+  list-style-type: disc;
+  -webkit-columns: 2;
+  -moz-columns: 2;
+  columns: 2;
+  column-gap: 25%;
+  padding-top: 1rem;
+}
+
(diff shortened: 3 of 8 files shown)


AemieJ:
This is just a basic draft on the creation of a sitemap so I have suggestions regarding designs and not design-related as well.
1. Within the Camel Website sitemap, The blog is displayed after the releases. So either we include all of the camel releases or we just included the releases within the blog category. I don't see the reason to include both of them. It's redundant. 

Another is within the community where support is included twice. This is only 'cause both Mailing Lists and Getting Help tends to have the same title. So @zregvart is it best to call the mailing list as the title instead of having support as the title followed by mailing lists. I would like to know your opinions on this one.

Within apache camel security, it's best to just include the `CVE-` part only.

2. For the user manual sitemap, I was thinking to include FAQ as a separate module of the sitemap itself. In addition to it, we don't require the Components, EIP within the user manual sitemap, it's good to remove the ones that are related to the camel components section. 

3. I like the detailing of the component sitemap listing but we will have to think of a manner in which we present the layout. If the above two non-design related issues are handled, I would also like the columns to be in a count of 3, covers less vertical space. 
An idea on how I want the normal list to be presented is in the image below. For the components, I will need to give a look into that. 
camel-sitemap-design

aashnajena:
For all three points, I just want to say that I want to keep things as automated as possible. 

- Since we're giving the user easy access to all links present, I don't see why we shouldn't include Releases. The pages under "Releases" are different from those under "Blog" and picking pages to remove from the sitemap would just make it more customized, less automated.  
- Renaming Mailing List is out of the scope of this PR, but we should definitely do that. We should also consider moving "building" and "sources" with the other documentation pages.
- Again, making FAQ into a separate module will mean moving away from automation. I'd prefer to keep this first iteration as simple as possible. 
- I did initially go with 3 columns because it definitely does use much less vertical space. But 3 columns become very congested for entries under FAQ or Security which have long headings, so I went with the safer middle option. 

Overall, since this is the first iteration, I want to keep it simple with one common layout and common CSS. It's hard to keep track of which titles to trim and which entries to exclude because the pages keep growing and changing.

zregvart:
> Another is within the community where support is included twice. This is only 'cause both Mailing Lists and Getting Help tends to have the same title. So @zregvart is it best to call the mailing list as the title instead of having support as the title followed by mailing lists. I would like to know your opinions on this one.

I'd rename the Mailing list to Mailing list.

zregvart:
I'm a bit confused I thought we wanted a single sitemap to include all website content, this introduces several sitemaps, one for Hugo content and one for each Antora component. Not sure how would that work with a single link from the footer?

I'm also not sure if we would like to list each page in the sitemap page, that might lead to really big sitemap page and lead to a lot of scrolling. I think we need to fit the sitemap within one to two page-downs. There's a [study on sitemap usability](https://www.nngroup.com/articles/site-map-usability/) that speaks to this somewhat.

AemieJ:
@zregvart  I read the article and within that, it stated **A site map is, after all, a map; it should not be a navigational challenge of its own.** which I agree too so it's good if we break down the whole sitemap into 2 pages instead of providing separate categories for each project but that will be possible if we include only the required necessities. In my opinion, blog lists are not required for the sitemap. Even for each specific project, only certain features such as traits or a list that is provided within each project is required. 

@aashnajena I think it's best to segregate what is required and what is not on the sitemap on the Jira issue before further implementation.

aashnajena:
@zregvart I thought the task was to create a user-readable organized version of the sitemap which we already have. Since our XML sitemaps have a sitemap-index followed by 8 sitemaps (that encompass all pages on the website), I tried to mimic that. I didn't realize that the expectation is much different here. In that case, we can discuss on the Jira issue and re-do this task again.

PR #434, opened 2020-07-21
feat: add github action for image optimization

AemieJ:
* The performance due to lack of image compression lowers for the webpage. Gulp Imagemin wasn't showing the results as expected thus I chose to use a GitHub action which is easy, effective, and does lossless compression.

* This will be triggered each time a PR is created. The compression for the current status of the branch is shown in the snippet of the image below. ( More files were edited than the ones shown in the image )
github-action

* However, I require clarification on whether the images from the **public** folder only need to be compressed or from the **static** folder as well?

### DIFF
--- a/.github/workflows/calibreapp-image-actions.yml
+++ b/.github/workflows/calibreapp-image-actions.yml
@@ -16,5 +16,5 @@ jobs:
           jpegProgressive: false
           pngQuality: "80"
           webpQuality: "80"
-          ignorePaths: "node_modules/**,build"
+          ignorePaths: "static"
           
--- a/.github/workflows/calibreapp-image-actions.yml
+++ b/.github/workflows/calibreapp-image-actions.yml
@@ -1,0 +1,20 @@
+name: Compress images
+on: pull_request
+jobs:
+  build:
+    name: calibreapp/image-actions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Compress Images
+        uses: calibreapp/image-actions@master
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/calibreapp-image-actions.yml
```diff
@@ -0,0 +1,20 @@
...
+
+      - name: Compress Images
+        uses: calibreapp/image-actions@master
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          jpegQuality: "80"
+          jpegProgressive: false
+          pngQuality: "80"
+          webpQuality: "80"
+          ignorePaths: "node_modules/**,build"
```
zregvart, line 19: I don't think we need this we don't have node_modules nor build directories. Would this include only `public`?

```suggestion
          ignorePaths: "!public"
```

zregvart:
>     * However, I require clarification on whether the images from the **public** folder only need to be compressed or from the **static** folder as well?

I think we can optimize only images that are in `public`, those will end up published on the website.

.github/workflows/calibreapp-image-actions.yml
AemieJ: As we don't need to include the images within the **static** folder, I think I will change the ignorePaths to static

zregvart:
Doesn't seem like the action [can commit](https://github.com/apache/camel-website/pull/434/checks#step:4:96) the changes.

AemieJ:
> Doesn't seem like the action [can commit](https://github.com/apache/camel-website/pull/434/checks#step:4:96) the changes.

I believe it's only when you merge this and a PR is then on a PR, the action will take place. 'cause when I took one of the branches and created a PR in AemieJ/image-optimization. Then it takes the action into account and optimizes the image files.

PR #440, opened 2020-07-27
Revert "CAMEL-15185: Footer design"

oscerd:
Reverts apache/camel-website#428

### DIFF
--- a/antora-ui-camel/src/css/footer.css
+++ b/antora-ui-camel/src/css/footer.css
@@ -19,46 +19,12 @@ footer .footer {
 
 footer .footer p {
   flex-basis: 100%;
-  margin: 3rem 0 0.5rem 0;
+  margin: 0.5rem 0;
   padding: 0 1rem;
 }
 
-footer .footer p.remark::after {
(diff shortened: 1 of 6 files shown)


oscerd:
Reverting it, because the website build seems to be broken.

AemieJ:
@oscerd when I do `yarn checks` for the footer-design I receive `page not found from components/3.4.x/eips/request-reply.html to community/support` however request-reply.html isn't present anymore within the website and the PR #428 preview works perfectly.

oscerd:
preview build doesn't do what the CI build does there is no checks for link there.

zregvart:
Yes, we don’t run checks on Netlify because running them would use up the free build minutes we have there.

zregvart:
@aemiej remove everything from the public directory and rebuild, we don’t have a way to remove just the deleted pages.

AemieJ:
@zregvart I did that still `yarn checks` failed for `integration.html` under **camel-k** and I observed that it is present in the other PR that has been merged and built as well.

PR #443, opened 2020-07-31
CAMEL-15360: fix HTML validation issues

zregvart:
This greatly simplifies the footer design and fixes HTML issues
identified by `check:html` script.

We're now using Unicode full width plus symbol so we don't need the SVG
icons for show/hide menu.

### DIFF
--- a/antora-ui-camel/src/css/footer.css
+++ b/antora-ui-camel/src/css/footer.css
@@ -62,11 +62,13 @@ footer .footer .context:last-child a::after {
 }
 
 footer .footer figure.logo {
+  display: flex;
   flex-direction: row;
   flex-grow: 1;
-  align-items: flex-start;
   justify-content: center;
   margin-inline-start: 0;
+  align-content: center;
+  align-items: center;
 }
 
 footer .footer figure.logo img {
@@ -91,7 +93,7 @@ footer .footer dl dt {
   margin-bottom: 1.5rem;
 }
 
-footer .footer dl dt::after {
+footer .footer dl dt label:first-of-type::after {
   content: '';
   display: block;
   position: relative;
@@ -100,10 +102,6 @@ footer .footer dl dt::after {
   height: 2px;
   width: 3rem;
   background: var(--footer-font-color);
-}
-
-footer .footer .footer-menu {
-  display: block;
 }
 
 footer .footer dl dd {
@@ -126,10 +124,9 @@ footer .footer dl dd {
   height: 2rem;
 }
 
-footer .footer .break-row-footer {
-  display: block;
+footer .resources {
   flex-basis: 100%;
-  height: 0;
+  text-align: center;
 }
 
 footer .footer .footer-icons {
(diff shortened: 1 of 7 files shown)


antora-ui-camel/src/css/footer.css
```diff
@@ -182,59 +181,45 @@ input[type="checkbox"] {
...
+    top: 2rem;
+    left: 45vw;
+    width: 1.75rem;
+    height: 1.75rem;
+    transition: 0.2s;
   }
 
-  footer .footer .footer-menu {
-    margin-top: 1rem;
+  footer .footer input[type="checkbox"] {
```
AemieJ, line 198: Is this meant to be changed as `footer .footer input[type="checkbox"]:checked` as it's the same as the stated class as above and thus the preview is failing for the same?

zregvart: Yeah I noticed a quirk artefact in Chrome so I fixed it last minute without running the builds/checks, should be okay now.

antora-ui-camel/src/css/footer.css
```diff
@@ -182,59 +181,45 @@ input[type="checkbox"] {
...
-
-  input[type="checkbox"]:checked + label img.show-menu {
-    display: none;
-  }
-
-  input[type="checkbox"]:checked ~ label img.hide-menu,
-  input[type="checkbox"]:checked ~ div.footer-menu {
-    display: block;
+  footer .footer input[type="checkbox"]:checked::before {
+    transform: rotate(45deg);
```
AemieJ, line 215: `transform: rotate(90deg);` 
This is 'cause even on clicking it rotates to a plus sign instead of a cross. To make it cross, the following code achieves it.

antora-ui-camel/src/css/footer.css
```diff
@@ -182,59 +181,45 @@ input[type="checkbox"] {
...
-    height: 35px;
-    margin-top: -2.5rem;
+  footer .footer input[type="checkbox"] {
+    display: inline;
+    position: relative;
+    top: 2rem;
+    left: 45vw;
+    width: 1.75rem;
+    height: 1.75rem;
+    transition: 0.2s;
```
AemieJ, line 195: In addition to this, there is an outline around the box, to prevent this following code achieves it: 
````
-webkit-appearance: none;
-moz-appearance: none;
-o-appearance: none;
appearance: none;
outline: 0;

File: antora-ui-camel/src/css/footer.css [hunk 4]
```diff
@@ -182,59 +181,45 @@ input[type="checkbox"] {
...
 
-  footer .footer .footer-menu {
-    margin-top: 1rem;
+  footer .footer input[type="checkbox"]:checked {
+    transform: rotate(45deg);
   }
 
-  input[type="checkbox"] + label img.show-menu {
+  footer .footer input[type="checkbox"]::before {
+    background-color: var(--footer-background);
```
AemieJ, line 203: This line of code isn't required with the above suggestions I made.

AemieJ:
outline-checkbox
@zregvart I found something odd about the checkbox outline so, with the following changes I suggested, the outline is removed.

antora-ui-camel/src/css/footer.css
zregvart: This is caused by both `input` and `::before` being rotated by 45deg. I've gone with a different implementation in the latest changes.

antora-ui-camel/src/css/footer.css
zregvart: `appearance` is a [experimental](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) [feature](https://caniuse.com/#feat=css-appearance)we can't rely on it for website to render on most devices. I've gone with a different implementation in the latest changes.

antora-ui-camel/src/css/footer.css
zregvart: I've gone with a different implementation in the latest changes.

zregvart:
> @zregvart I found something odd about the checkbox outline so, with the following changes I suggested, the outline is removed.

This is the input showing beneath the content made with `:before`. I've implemented a different solution in the latest changes. We should not remove outline, it helps folk what navigate using the keyboard. That being said I found no way to have the checkbox focused, in the right place with keyboard navigation and not visible at the same time in the latest changes; so we don't have keyboard focus indication (outline) in the latest changes for the plus sign or for the section headings (`dt`).

layouts/partials/footer.html
```diff
@@ -6,66 +6,50 @@
...
-                <input type="checkbox" id="show-overview"/>
-                <label for="show-overview"><img class="show-menu" src="/_/img/show-footer-menu.svg"></label>
-                <label for="show-overview"><img class="hide-menu" src="/_/img/hide-footer-menu.svg"></label>
-                <div class="footer-menu">
-                    <dd><a href="/blog/">Blog</a></dd>
-                    <dd><a href="/docs/">Documentation</a></dd>
-                    <dd><a href="/community/support/">Community</a></dd>
-                    <dd><a href="/blog/">Download</a></dd>
-                </div>
+                <dt><label for="footer-toggle-overview">Overview</label><label for="footer-toggle-overview">&#65291;</label></dt>
```
AemieJ, line 15: `<label for="footer-toggle-overview">&#65291;</label>`* is better included with a separate unique **id** to distinguish as it appears for desktop screen as well and the same case for each label with `&#65291;` in content.
- `<label for="footer-toggle-overview" id="expand">&#65291;</label>`

AemieJ:
@zregvart yes, I observed that it's better just the **+** is included now within the desktop version (> 1024px) as well so best we include separate id for those labels and change **display** property as per that.

layouts/partials/footer.html
zregvart: Yeah I was preoccupied with mobile version that I forgot to check the desktop one. Not sure if assigning `id`s here would be the correct approach. Assigning same `id` to more than one element is not valid HTML. I've set the `display: none` via CSS selector: `footer .footer dl dt label:nth-child(2)`.

zregvart:
@AemieJ I think this is now okay to get merged, thanks for reviewing, can you check the latest version when the preview is built?

AemieJ:
LGTM.

PR #444, opened 2020-07-31
CAMEL-15335: update the search layout design

AemieJ:
### Search Layout Design 
* Version is separated from the title of the section to keep a neater look and easier to understand the search result.
* The borders have been removed as they don't fit in with the theme of the website. 
* On hovering a particular link, the whole section will give a shade of #f5f5f5 providing a highlighting effect.
* Algolia attribute has been at the bottom of the search result. 

search-result

* Changed the look for **Nothing found** for no results found on a search query.

no-result


antora-ui-camel/src/css/header.css
```diff
@@ -375,13 +375,33 @@ body {
 
 #search_results {
   top: 3.38rem;
-  background: #fffe;
+  background: #fff;
```
zregvart, line 378:
```suggestion
  background: var(--navbar-background);
```

antora-ui-camel/src/css/header.css
```diff
@@ -412,9 +462,22 @@ body {
   padding-top: 0.3rem;
 }
 
+#search_results div.footer-search {
+  display: inline-flex;
+  float: right;
+  margin-right: 0.75rem;
```
zregvart, line 468: `float` doesn't make sense in flex layout, better to use padding vs margin so not to introduce scrollbars

```suggestion
  padding: .75rem;
  width: 100%;
```

antora-ui-camel/src/css/header.css
```diff
@@ -390,15 +410,45 @@ body {
...
+}
+
+#search_results div.result:hover {
+  background: var(--color-smoke-50);
+}
+
+#search_results div.section {
+  display: flex;
+}
+
```
zregvart, line 425: This was causing scrollbars to show

```suggestion

  #search_results dl {
    margin: 0;
  }
```

aashnajena:
This implementation looks very neat and I like how the versions are displayed.

I have one doubt - is it common to highlight the entire section when hovering over one search result? It's a little confusing as to what I'm clicking - one search result or the entire section (even with the underline detail). I think it would be much clearer if we just highlight the search result user is hovering on, or at least provide some clear indication of what the user is clicking on (the background grey and underline grey are very similar, and on Chrome, only the heading gets underlined). 

Another observation is that when you move from one search result to its next within the same section, the grey background flickers because there's some space between two results which does not have the same hover property.

aashnajena:
This PR needs to wait. I'm observing that [this additional commit](https://github.com/algolia/docsearch-configs/commit/77634d58dab0f46f7dfc020a7b7fae1e9b390fe2#diff-a63c497e1db6a7dcc07ef35536d19967) made by an Algolia maintainer on our config file is causing some problems. @zregvart I have opened a [PR on the Algolia repository](https://github.com/algolia/docsearch-configs/pull/2133) to fix these issues. 

image

AemieJ:
@aashnajena yes, I already observed it. I am currently focusing on dealing with it without making change to the config file.

AemieJ:
@zregvart this is why I suggested pagination for search results within the JIRA issue 'cause even if I show 5 of the results for some cases, it's going to overflow and the only time the scrollbar doesn't appear is when 3 results are presented on the search result but in my opinion, 3 results are not a convenient or suitable number to show to the user for a search query.

aashnajena:
@AemieJ notice that blogs are not appearing as they should. They are getting grouped by categories. I don't think we should update the design layout without first fixing the flaws on the config side, it would be a little hacky. PR #445 along with PR #2133 on the Algolia repo will fix the underlying issue, and then we can work on top of that to improve the layout.

Also, since different versions of the same page appear as multiple search results, shouldn't we be showing more search results to make sure we are including all relevant entries? Sorry if I'm missing out on something that's already been discussed. I don't understand what's the issue with the current format.

AemieJ:
@aashnajena I am only talking about the default value being Documentation problem. What do you exactly mean by blog not being display as it is? Do the blogs need to be represented as -
```
Blog 
Category
Title
```
If it's the case as above then the hierarchy needs to be reordered in the case of blogs.

aashnajena:
The difference between current and expected behaviour is visible on the snapshot on PR #445

antora-ui-camel/src/css/header.css
AemieJ: This doesn't remove scrollbars necessarily. Scrollbars only appear when the text covered by dl has height > 80vh which occurs in a few cases. Hence, I suggested using pagination to avoid scrollbars usage. However, for this PR, I wanted to just have a basic implementation for the search design layout.

zregvart:
Do I understand correctly we need #445 merged and base this effort on that?

AemieJ:
@zregvart in the algolia doc-search, Aashna created a PR, if that is merged and PR #445 is merged as well, then I can continue working on this PR.

package.json
```diff
@@ -8,7 +8,6 @@
     "build:sitemap": "gulp sitemap",
     "build": "run-s build:*",
     "preview": "run-s preview:*",
-    "preview:antora": "antora --clean --fetch antora-playbook.yml",
```
zregvart, line 11: I don't think this was meant to be included here.

antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
```diff
@@ -46,22 +46,22 @@
...
+              if (hit.hierarchy.lvl6 !== null) section = section + '/' + hit.hierarchy.lvl6
               var breadcrumbs = Object.values(hit.hierarchy)
                 .slice(1)
                 .filter((lvl) => lvl !== null)
-                .join(' &raquo; ')
+                .join(' / ')
 
               d.breadcrumbs = ((breadcrumbs !== '') ? breadcrumbs : section)
-              d.snippet = hit._snippetResult.content.value + '...'
+              d.snippet = hit._snippetResult.content.value.split('&quot;').join('') + '...'
```
zregvart, line 64: Seems that `hit._snippetResult` can be undefined, are we using `d.snippet`?

AemieJ: Yes, that makes sense, I altered the code to take care of it so it doesn't split for undefined values. Yes, we use `d.snippet` within the `summary` tag.

aashnajena:
I think the cancel icon's position needs to be fixed for smaller screens.

AemieJ:
@aashnajena I will look into that and fix it.

antora-ui-camel/src/css/header.css
```diff
@@ -356,7 +418,7 @@ body {
   }
 
   #search-cancel {
-    bottom: calc(50% - 0.45rem);
+    margin-top: 0.75rem;
   }
 
   #search_results {
```
zregvart, line 424: I think it looks a bit better if we take up full width on smaller screens
```suggestion
  #search_results {
    left: 0;
    width: 100vw;
```

zregvart:
Thanks!

PR #445, opened 2020-08-02
fix: search results

aashnajena:
The search results for blogs, articles, books, download are showing up like this:

image

The expected behaviour is shown below. I've made [PR #2133](https://github.com/algolia/docsearch-configs/pull/2133) on the Algolia repo to fix changes in the config file. With the changes in the config file + changes in this PR, we will be able to achieve the expected behaviour. 

image

### DIFF
--- a/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
+++ b/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
@@ -53,10 +53,10 @@
             const data = hits.reduce((data, hit) => {
               const d = {}
               d.url = hit.url
-              var section = hit.hierarchy.lvl1
-              if (hit.hierarchy.lvl0 !== null) section = section + ' [' + hit.hierarchy.lvl0 + ']'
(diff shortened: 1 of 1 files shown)


zregvart:
Do we need https://github.com/algolia/docsearch-configs/pull/2133 merged before this, or can we merge this first?

aashnajena:
We need the Algolia PR to be merged first

aashnajena:
This PR can be merged now.

AemieJ:
I checked with the preview even if blogs, downloads, articles show accurately. We certainly don't want this layout for the project documentation, it's confusing. Why is it that in spite of the doc search config being correct and your code corresponding to it is correct, for documentation is a different layout than expected?  

@zregvart is it 'cause the site hasn't been built yet or any other reason?
search-result-now

zregvart:
I’ve triggered a rebuild of the website. I’ll need to take a look at the build configuration, doesn’t seem to trigger automatically on changes.

PR #448, opened 2020-08-04
Improvize on the style of navbar and navbar items layout

AemieJ:
This PR deals with the following changes in the layout: 
1. For the Desktop ( > = 1024px)
* Remove the border-bottom and apply a mild box-shadow[1]
* Replace the color change on hover to the navbar items with a border-bottom on hover to the navbar items.
* Increase the size of navbar tools such that the navbar looks consistent. 
> [1] the box-shadow appears only when scrolled through the webpage. 

2. For the smaller screen width ( < 1024px)
* In here, re-structured the navbar cross button. 
* Based on designs suggested on the [Jira Issue](https://issues.apache.org/jira/browse/CAMEL-15140?jql=project%20%3D%20CAMEL%20AND%20component%20%3D%20website), this [mockup](https://issues.apache.org/jira/secure/attachment/13007475/Mockup-7.png) has been implemented.

3. Extra Changes 
* Remove the **About** section from the navbar ( as well present in the footer )
* Remove any link related to the dropdown within the navbar.

PR #451, opened 2020-08-06
CAMEL-15140: Improve Mobile Main-Menu

aashnajena:
In this PR I have :
- Removed the "About" menu-item from the header, because it now present in the footer
- Changed header.hbs and header.html layout files to remove code for sub-menus
- Removed styling for submenus (.navbar-dropdown class is no longer required)
- Changed the design of the main menu on smaller screens as discussed on the JIRA issue.

Here's how it looks for smaller screens in landscape and portrait mode respectively :

image

image


AemieJ:
@aashnajena PR #448 deals with those changes along with the desktop version as well.

aashnajena:
Thanks for pointing out, I didn't realise you're working on this issue. Please link it to the JIRA issue so that it's easier to track. Since I've already implemented it now, maybe we can take some implementation bits from here and use them eg. the images, the resizable menu (so that it doesn't go to scroll even if orientation is landscape), the dynamic sizes of menu-items which change with height & width of screen, and having paths of logo-images as part of config.toml, so that they are easy to map, if we want to change the images.

zregvart:
@aashnajena having merged #448, is there something that we miss from this or can this PR be closed?

aashnajena:
This can be closed now

PR #452, opened 2020-08-08
CAMEL-15371: underline links

AemieJ:


antora-ui-camel/src/css/doc.css
```diff
@@ -149,6 +149,16 @@
 .doc a {
   color: var(--link-font-color);
   word-break: break-word;
+  text-decoration: underline;
+  text-underline-position: under;
```
zregvart, line 153: This bit looks a bit weird to me.

```suggestion
```

AemieJ: You require this particular line so that there is a minimum distance between the underline and the text.

zregvart: Right, so this is with `text-underline-position: under`:

Screenshot_2020-08-11 Home - Apache Camel

and without:

Screenshot_2020-08-11 Home - Apache Camel(1)

I think what you're reacting to is that the underline is quite noticeable, perhaps adding to make it lighter:

```suggestion
  text-decoration-thickness: 1px;
  text-decoration-style: dotted;
```

Would help, this is how it renders in Firefox:

Screenshot_2020-08-11 Home - Apache Camel(3)

AemieJ: I am not sure about to dotted style. It would look quite odd. I will look into the thickness.

AemieJ: Thickness feature isn't present for chrome browser and the dotted style looks inconsistent. Either we just not include under or we include it. It's best to keep it simple. Or I could use the `border-bottom` property.

zregvart:
Not sure about the underline being below the descenders. My personal preference would be

```css
text-decoration: underline;
text-decoration-style: dotted;
```

There are some methods on different underline styles at [css tricks](https://css-tricks.com/styling-underlines-web/) and I remember reading about the underline journey at [medium](https://medium.design/crafting-link-underlines-on-medium-7c03a9274f9) (for inspiration).

@aashnajena and @delawen any thoughts on this?

AemieJ:
@zregvart alright I got your concern about it but with those changes, it looks messy in chrome browser, I found a method using background-image and linear-gradient on the CSS tricks site. 

```
background-image: linear-gradient(to right, #583ac2 50%, transparent 50%);
background-position: 0 1.1em;
background-repeat: repeat-x;
background-size: 7px 1px;
```

If it's okay with you, I will implement it in this manner, it will create a neater effect of underline on each browser. 
underline-link

zregvart:
> ```
> background-image: linear-gradient(to right, #583ac2 50%, transparent 50%);
> background-position: 0 1.1em;
> background-repeat: repeat-x;
> background-size: 7px 1px;
> ```

Does the `background-position` track different font sizes correctly? Might be a bit of a headache to have this properly positioned for all font sizes.

AemieJ:
@zregvart I have checked it, the `background-position` tracks all the font-size of the text and that of browsers correctly.

zregvart:
Thanks @AemieJ

PR #442, opened 2020-07-30
CAMEL-15300: Update Front Page Design

aashnajena:
Changed the front page design for header, functionalities and projects section. I have implemented the design for all three- desktop, tablet and mobile (this may not be perfect as of now) view. As of now, I am featuring the 3 latest blogs from "releases" on the front page. All icons/images are self-made.

I remember seeing a "What is Camel?" page somewhere on the website, but I can't find it now. Can someone point me to that page?

PR #370, opened 2020-05-23
CAMEL-14994: Improve Projects Page

aashnajena:
First draft for improving the projects page. I'm opening the PR so that the preview is available. Need feedback on a couple of things :
- I've taken icons from Eva, from where icons in the "Packed with Functionality"  section (on front page) are taken. Do the outlined icons look better or the filled ones?
- Do we need to add text explaining the icons?
- As pointed out on the JIRA issue, the mailing list is common for all projects, so should we include that in each project box?
- As pointed out on the JIRA issue, we should be adding links to examples. What's the best way to add those?
- Quarkus has a separate Gitter chat. Do other projects have this too? We can add links to those.

Also, in <1024 screens, the h1 headings were becoming smaller than other headings, so I removed that.

PR #453, opened 2020-08-08
CAMEL-15392: provide a modified neat card layout for community & docs

AemieJ:
### Layout Design Change 
The card layout design for the documentation and community page seemed plain compared to the other pages of the website. Thus, included rounded background over the links for the pages and rounded border over the section to provide the card-like layout. This will fit with the new design for the frontpage in PR #442 as well.

#### Community Page
contributing-design-card

### Documentation Page 
community-design-card


aashnajena:
This looks fine on the community page, I'd prefer if we make the button's width just enough to fit the content. For docs, we have three different links per card, and putting them all in the same button makes it look like one combined link. We can have 3 separate buttons (there's an image on #370 showing how that might look). But we might have to think about how it will look on smaller screens because 3 buttons take up a lot of space.

aashnajena:
For community page, this looks better to me.

image

AemieJ:
@aashnajena I agree with the change of width with respect to the context itself, I will make those changes. 

Other than that, for the documentation page, I temporarily included them within one button as I feel buttons with the text would be highly clustered. For that, I was planning on including icons on it, and a custom tooltip on hover but I needed opinions on that change.

PR #457, opened 2020-08-13
Emulate the parts of FontAwesome that we use

ppalaga:
Motivated by https://github.com/apache/camel-quarkus/issues/1410 , based on https://gitlab.com/antora/antora-ui-default/-/merge_requests/121/diffs

### DIFF
--- a/antora-ui-camel/src/css/fa-emulated.css
+++ b/antora-ui-camel/src/css/fa-emulated.css
@@ -1,0 +1,9 @@
+/* For perf reasons, we emulate only those few parts of Font Awesome css that we use without actually pulling the FA font */
+
+.fa {
+  display: inline-block;
+}
+
+.fa-lock::before {
+  content: "🔒";
+}
--- a/antora-ui-camel/src/css/site.css
+++ b/antora-ui-camel/src/css/site.css
@@ -1,5 +1,6 @@
 @import 'typeface-open-sans.css';
 @import 'typeface-droid-sans-mono.css';
+@import 'fa-emulated.css';
 @import 'vars.css';
 @import 'base.css';
 @import 'body.css';


zregvart:
I suspect that this will add significant weight to the pages, and I'm not sure if we're going to use more than several FontAwesome icons. Perhaps it would be worthwhile to investigate a [customized version](https://github.com/FortAwesome/Font-Awesome/wiki/Customize-Font-Awesome#the-easy-way) including only icons that we do use. As an alternative we could use the unicode lock symbol: `&#x1f512;` (&#x1f512;).

antora-ui-camel/gulp.d/tasks/build.js
```diff
@@ -46,7 +46,18 @@ module.exports = (src, dest, preview) => () => {
           const basename = ospath.basename(abspath)
           const destpath = ospath.join(dest, 'font', basename)
           if (!fs.pathExistsSync(destpath)) fs.copySync(abspath, destpath)
-          return path.join('..', 'font', basename)
+          return path.join('..', 'font', basename) + asset.search + asset.hash
```
zregvart, line 49: Not sure about this change, doesn't seem to be related to adding Font Awesome.

ppalaga:
Thanks for the feedback, @zregvart ! I am a frontend noob happy to follow any advice :) OK, let me try to extract only the relevant CSS and try the unicode glyph.

ppalaga:
c0e2e6b adopts the proposal to emulate `fa-lock` with 🔒 unicode character

ppalaga:
ups, there are changes in c0e2e6b that should not be there, let me filter them out

ppalaga:
f25b8f5 is what it should be

ppalaga:
The lock sign can be seen e.g. here https://deploy-preview-457--camel.netlify.app/camel-quarkus/latest/extensions/microprofile-metrics.html#_additional_camel_quarkus_configuration 
Would please somebody review?

zregvart:
@ppalaga this can be done without adding any new CSS, i.e. by adding the icon as [unicode character](https://docs.antora.org/antora/2.3/asciidoc/special-characters-and-symbols/) to the adoc file. I think that way we would not couple what icons can be used in docs to the state of CSS on the website.

ppalaga:
> @ppalaga this can be done without adding any new CSS, i.e. by adding the icon as [unicode character]

In our situation, it is rather hard to do, because we get `icon:lock[title=Fixed at build time]` hard coded from Quarkus tooling https://github.com/quarkusio/quarkus/blob/master/core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java#L78 and I'd prefer not to rewrite it.

I thought, because font awesome is commonly used in both AsciiDoc [1][2] and Antora (it is a part of the default UI now [3]), having some level of CSS support on our website would be quite natural too. 

[1] https://asciidoctor.org/docs/user-manual/#admonition-icons
[2] https://asciidoctor.org/docs/user-manual/#inline-icons
[3] https://gitlab.com/antora/antora-ui-default/-/merge_requests/121

zregvart:
> > 
> 
> In our situation, it is rather hard to do, because we get `icon:lock[title=Fixed at build time]` hard coded from Quarkus tooling [https://github.com/quarkusio/quarkus/blob/master/core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java#L78](https://github.com/quarkusio/quarkus/blob/master/core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java?rgh-link-date=2020-08-18T08%3A27%3A36Z#L78) and I'd prefer not to rewrite it.

Right, I thought we were in full control of the docs.

PR #455, opened 2020-08-11
fix: broken links and HTML validation

zregvart:
This fixes broken link to component reference and outstanding HTML
validation issues.

PR #458, opened 2020-08-13
Disable HTML validation temporarily

ppalaga:
@zregvart the links pass now, but the HTML validation is still failing. Is it OK to disable it so that we can get the Camel Quarkus 1.0.0 announcement out?


davsclaus:
+1 for me, create a JIRA about its disabled and point to this PR so we wont forget and can turn it back on, and fix the validation errors

zregvart:
With #455 merged we no longer need this.

ppalaga:
It looked like #455 needs more time, therefore this attempt for a minimal viable fix. Thanks @zregvart for finishing #455 that fast!

PR #459, opened 2020-08-13
CAMEL-15337: Add TOC to static content

zregvart:
Based on work by @AemieJ in #446 this simplifies the implementation by
relying on the implementation already provided by Antora.

The main issue with reusing the Antora implementation was that the
markup of rendered Asciidoc contains `<div class="sect?">` elements
around the chapters. Hugo rendered, static, content does not have this,
so the logic for the _on this page_ functionality needed to be adjusted
slightly because of this.

### DIFF
--- a/antora-ui-camel/src/css/static.css
+++ b/antora-ui-camel/src/css/static.css
@@ -1,6 +1,11 @@
 .static {
   margin: var(--static-margin);
-  max-width: var(--static-max-width--desktop);
+}
+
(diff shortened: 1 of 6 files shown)


zregvart:
@AemieJ can you have a look at this?

AemieJ:
@zregvart I didn't see that the Hugo missed that section. This is a much easier implementation and it works perfectly.

PR #460, opened 2020-08-13
fix: prevent zoom on search focus

zregvart:
For `input` elements with `font-size` less than `16px` Safari on iOS
zooms in on the element. This increases the font-size of the search
input and adjusts it's width, and positioning of the search cancel icon.


zregvart:
I think the search cancel icon can be a bit bigger and could be centered a bit better. I'll improve on this a bit later. I've also noticed that the "Search by Algolia" appears quite huge on mobile.

zregvart:
To keep it simple I'll do new PRs for other issues.

PR #462, opened 2020-08-15
CAMEL-15296:  Create sitemap for Camel website

aashnajena:
This PR adds the first version of sitemap for camel-website. Sitemap can be accessed from the footer. The problem with fetching links programmatically is that links like "misc" and "news" show up, even though they don't have any landing pages. We could choose to hard code all our links instead to avoid such behavior.

### DIFF
--- a/antora-ui-camel/src/css/site.css
+++ b/antora-ui-camel/src/css/site.css
@@ -23,3 +23,4 @@
 @import 'misc.css';
 @import 'community.css';
 @import 'docs.css';
+@import 'sitemap.css';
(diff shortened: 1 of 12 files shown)


AemieJ:
I like the basic design layout but I just want to clarify that we are going to include more links right? 'cause almost all the links present on the sitemap are there on the footer as well.

aashnajena:
Please suggest which other pages we should put on the sitemap, and I can add those..

aashnajena:
I think we can merge this and improve on it later.

AemieJ:
I believe that we need certain elobaration in site map upto level 1 for the sub projects as well. 'cause if I had to comparatively look at the site map of Apple, it is quite useful as it provides me the direct information neither the navbar nor footer provides. So we could incoperare a more elobrated site map upto max two pages.

However I don't mind this being merged, I can work on it as I have a few additional ideas which I would like to improve on later.

zregvart:
Thanks!

PR #466, opened 2020-08-18
CAMEL-15421: render of AsciiDoc admonitions for Antora

AemieJ:
The AsciiDoc admonitions weren't being rendered on the antora pages. Hence, I made a few tweaks to the CSS for the admonitions to be rendered. 

admonition-block-antora

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -313,7 +313,7 @@
 }
 
 .doc .admonitionblock .icon {
-  position: absolute;
+  position: relative;
   top: 0;
   left: 0;
   font-size: calc(15 / var(--rem-base) * 1rem);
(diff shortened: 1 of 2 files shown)


zregvart:
Awesome, thanks! Perhaps we could tweak the background coluor a bit, that green is a bit off brand for Camel. It can be changed here:

https://github.com/apache/camel-website/blob/c86e67a1b072cd004325a1b7ff80a4d537a97345/antora-ui-camel/src/css/vars.css#L87

Perhaps a lighter version of the ASF blue would suit us better?

zregvart:
I've fiddled with this a bit, I think it looks a bit nicer without the rounded corners:

Screenshot_2020-08-18 Camel Quarkus extensions reference Apache Camel(1)

zregvart (approved):

zregvart:
Thanks!

PR #469, opened 2020-08-19
Provide new layout for Documentation page

aashnajena:
Opening this PR so that everyone can have a preview of what I was suggesting on [Camel-15392](https://issues.apache.org/jira/browse/CAMEL-15392). Benefits of new layout :
- Takes less vertical space
- Uses dark orange and light grey buttons. This draws attention to the "documentation" link as we wanted. Current layout uses buttons with partitions and a florescent orange color, which are not used anywhere else on the website. Personally the partitions make the buttons look like it has two parts, whereas they're both the same link, so it doesn't make much sense. For example, a partitioned button on [this page](https://getfedora.org/en/workstation/download/) provides a download link for either windows or macOS.
- I think this color scheme is a lot more elegant, since there aren't too many orange buttons in one line. It doesn't split for >1024px break-point as well. 
- It draws structure from the way our front page is designed. Earlier we had a outlined box layout for "functionalities" as well as "projects" section on the front page, so the box layout for docs made sense. I think it's logical to follow the same kind of structure everywhere.
- For <1024px the layout is pretty much the same as before, just without the border.


zregvart:
The build of Hugo pages is failing with:

```
Error: Error building site: "/opt/build/repo/content/docs/_index.md:69:6": failed to extract shortcode: shortcode "div" has no .Inner, yet a closing tag was provided
```

And we seem to have more changes that caused more conflicts here.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -74,29 +74,30 @@ header.frontpage p {
...
 }
 
-.frontpage a.button-dark,
-.frontpage a.button-dark:hover {
+a.button-dark,
+section.frontpage a.button-dark,
+section.frontpage a.button-dark:hover,
+a.button-dark:hover {
   white-space: nowrap;
   padding: 0.7rem 1rem;
```
zregvart, line 82:
```suggestion
  padding: 0.3rem 1rem;
```

I think smaller buttons look a bit less imposing:

Screenshot_2020-08-24 Documentation - Apache Camel

antora-ui-camel/src/css/frontpage.css
```diff
@@ -74,29 +74,30 @@ header.frontpage p {
...
   outline: none;
   display: inline-block;
-  box-shadow: 0 4px #8e480b;
+  background-image: none;
 }
 
-.frontpage a.button-light {
+a.button-light,
+a.button-light:hover {
   padding: 0.7rem 1rem;
```
zregvart, line 95:
```suggestion
  padding: 0.3rem 1rem;
```

We could try a bit smaller buttons.

antora-ui-camel/src/css/frontpage.css
```diff
@@ -74,29 +74,30 @@ header.frontpage p {
...
   margin-right: 0.5rem;
   border-radius: 3rem;
   outline: none;
   display: inline-block;
-  box-shadow: 0 4px #8e480b;
+  background-image: none;
 }
 
-.frontpage a.button-light {
+a.button-light,
```
zregvart, line 93: I'd start with a `a.button` selector and then have a `a.button.light` and a `a.button.dark` variant, that way any change to the `a.button` doesn't have to be redone for both variants.

antora-ui-camel/src/css/docs.css
```diff
@@ -1,154 +1,81 @@
...
+  width: 70%;
 }
 
-.camel-project .camel-documentation .links a {
-  color: var(--navbar-background);
-  font-weight: bolder;
+.docs .box img,
+.community .box img {
+  margin-top: 4rem;
+  height: 8vw;
```
zregvart, line 23: We might need to have the width instead of the height here, or redo the SVGs so they're of the same aspect ratio.

Screenshot_2020-08-24 Community - Apache Camel

zregvart (commented):
:+1: looks good to me, few improvements that I can think of. I guess we need to rebase this.

antora-ui-camel/src/css/docs.css
aashnajena: When exactly are you seeing this problem? For me, the alignment is correct for all screen widths. We can still add a max-width rule to make sure this doesn't happen.

aashnajena: This should be solved now.

zregvart:
I think with #471 merged we no longer need this. @aashnajena please reopen if there is something missing from this work.

PR #476, opened 2020-09-07
Fix Map generic types for @Configurer

apupier:
not sure about this one but the Map<Country> doesn't sound possible


apupier:
cc @oscerd @davsclaus

davsclaus (approved):

davsclaus (approved):

PR #500, opened 2020-12-01
chore: rework download and release pages

zregvart:
The download page, release archive and release pages now include
sub-projects. For Camel releases three kinds of releases are listed on
the downloads page: LTS, latest and legacy (for 2.x). Archetypes for
creating release notes which drive this content were added, existing
release notes were updated and new release notes were added for latest
releases of sub-projects.


oscerd (approved):

oscerd:
I'm fixing the docs in ckc, so the website will build

oscerd:
Fixed the links on ckc.

zregvart:
Let me highlight some of the changes.

1. The [downloads page](https://deploy-preview-500--camel.netlify.app/download/) now lists releases from subprojects and LTS, latest and legacy releases for Camel, there is also a blurb at the top to explain what LTS is. I would appreciate someone looking over that text
2. From the downloads page versions link to individual releases, for example [Camel 3.4.4](https://deploy-preview-500--camel.netlify.app/releases/release-3.4.4/) or [Camel-K](https://deploy-preview-500--camel.netlify.app/releases/k-1.2.1/). Releases for sub-projects are on `/release/{short name}-{version}/` URLs (we can't change the existing release URLs)
3. Release pages have issues linked, we had this for Camel, now for sub-projects this is pulled from GitHub, with nuances as different sub-projects do this differently
4. Link to download artifacts now got to ASF page (they're not direct), this is to help with artifacts that are deleted from dist as that half-way page will offer a link to the archives

I haven't devoted much time to designing this, more-or-less it's the same design we already had, I wanted to get the data to render first. I'd welcome design suggestions but I won't be incorporating those, bar major flaws, in this PR. This can be polished in subsequent PRs.

zregvart:
I see some issues in rendering Camel K Runtime download and HTML issues, I'll fix those now.

zregvart:
I think we're good now, turned out I didn't remember to remove `draft: true` and that I didn't close property a HTML tag. I'll keep this around for folk to comment before I merge this, probably tomorrow.

davsclaus (approved):

FrankyGee:
Looks nice and clean, although I have never used the original :)

A couple thoughts about the content:
* First paragraph where it mentions breaking changes: a link to the upgrade guide would be handy https://camel.apache.org/manual/latest/camel-3x-upgrade-guide.html
* First paragraph has some awkward grammar: "...even though we strive to be as backward compatible as possible, but we do, on occasion, introduce...". "even though" & "but" is a double negative, which doesn't really make sense. You can delete either one and it should read cleanly.
* Third paragraph where it mentions "Please do report any issues...": a link would be handy. Perhaps https://camel.apache.org/community/support/

zregvart:
I've reworded the text at the top of the downloads page and added links to further documentation, fixed broken links for camel-k-runtime and added month/date of each release.

The download links should be compliant with the [requirements by infra](https://infra.apache.org/release-download-pages.html), current downloads go through the `https://www.apache.org/dyn/closer.lua/` and older releases go to `https://archive.apache.org/dist/`. By mistake links were not over TLS, now that's fixed.

Change in: https://github.com/apache/camel-website/compare/65ad94d4ecfa31d2fe87a89e99432daeecf52dc7..586762c557a756a32d682f897b0662c3f9166b78

zregvart:
Preview is [live](https://deploy-preview-500--camel.netlify.app/download/) (same url).

zregvart:
I think this is more or less everything but a couple of things:

1. perhaps it would make sense to make use `kind` instead of arbitrary filters in release note front matter I'll look into that and see if it simplifies the implementation
1. handling of two kinds, for example two overlapping LTS releases, not sure how to handle that. One idea might be to add `last: N` support in the `data/release-categories.yaml`, i.e. `last: 2` on a LTS will show last two LTS releases, but that assumes that there will be always two overlapping LTS releases, and I'm not sure that's true

zregvart:
I went with 1., so front matter for release notes now has `kind`, IMHO this makes it easier to figure out on the release note, i.e. just set `kind: lts` or `kind: legacy`. I've also added the SHA1 cut-off date to switch between sha512 and sha1 hash files.

Still not sure how to handle two overlapping LTS releases, do we feature both (say 3.4.x and 3.7.x) or just the latest LTS; and how many overlapping LTS releases there can be, i.e. for some time 3.4.x and 3.7.x might overlap, and after a while 3.4.x is out of support so only 3.7.x can be shown. With the code as is now we'll feature only the latest LTS release.

davsclaus:
We plan on releasing 2 LTS per year so there will be overlapping, eg 3.4 and 3.7, and then later 3.7 and 3.10 and so on. I think all active LTS should be on the download page, eg we will do 3.4.5, 3.4.6, 3.47 while we do 3.7.0, 3.7.1, 3.7.2 and so on.

Great work on the download page, nice to see a single page with all of them. 

We may want to add a note that Camel Spring Boot and Camel Karaf are part of the Apache Camel release.

zregvart:
@davsclaus I'll make it fixed last 2 LTS releases then, perhaps using that `last: N` logic in the `data/release-categories.yaml`. I've filed [CAMEL-15910](https://issues.apache.org/jira/browse/CAMEL-15910) to add Spring Boot and Karaf release artifacts, then we can add them to the table with the release.

zregvart:
Okay, I think I've driven this as far as it can be driven. We now have:

- support for last 2 LTS releases (not shown currently as there is only one LTS release)
- sha512 vs sha1 pivot on 2018-11-10
- anchor links to headings on download/archived releases pages
- lots of comments to help with maintenance

Please have another look, I'll leave this open until tomorrow, unless serious issues are found by then I'll merge.

PR #505, opened 2020-12-09
Added camel-kafka-connector Binaries releases.

valdar:
The intend is to create a Binaries section for camel kafka connector in the download page auto magically linking to the content ASF svn

### DIFF
--- a/data/release-categories.yaml
+++ b/data/release-categories.yaml
@@ -87,6 +87,9 @@ camel-kafka-connector:
     - name: 'Sources'
       path_format: 'camel/camel-kafka-connector/{version}/apache-camel-kafka-connector-{version}-src.zip'
       filename_format: 'apache-camel-kafka-connector-{version}-src.zip'
+    - name: 'Binaries'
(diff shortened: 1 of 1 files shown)


oscerd (approved):

zregvart:
Thanks, this will do the desired effect, but we also need binaries for 0.6.1 uploaded to dist, otherwise we'll serve a broken link.

Either that or we need to modify the template and add make this conditional. For example we do this for sha1 vs sha512 hash files:

https://github.com/apache/camel-website/blob/03a69996594433ecdd078f6f4384f56f562acec9/layouts/partials/releases/downloads.html#L1-L2

https://github.com/apache/camel-website/blob/03a69996594433ecdd078f6f4384f56f562acec9/layouts/partials/releases/downloads.html#L125-L128

A bit more _proper_ way of making this conditional would be to add the condition to the `release-categories.yaml`, i.e. something like `from: 2021-01-01`, and choose in the template based on that and the release date. This would then work for the sha1/sha512 and for future cases when we change something about the dist.

davsclaus (approved):

valdar:
Closing as no more needed see: https://issues.apache.org/jira/browse/INFRA-21181

PR #512, opened 2020-12-18
chore: add custom link as an download option

zregvart:
This adds custom link as an option for downloads. This way we can link
to binary downloads hosted elsewhere, i.e. other than the ASF list, but
still most likely on the camel.apache.org website.

Even though it supports version substitution via `{version}` variable,
for Camel Kafka Connectors it links to the documentation and we only
publish the "latest" version. There is also an issue that a release
version, say `1.2.3` doesn't match up with the documentation version, as
we do it today, i.e. we would have documentation version of `1.2.x`. The
solution might be to publish documentation for each version, but that
will take a toll on the website build times, and we can't really have
docs for all versions ever released, the website would be become really
unmanageable.

So this is a first pass, just to help users find the downloads, and we
need to figure what we would like to do WRT versioning.

### DIFF
--- a/data/release-categories.yaml
+++ b/data/release-categories.yaml
@@ -87,6 +87,9 @@ camel-kafka-connector:
     - name: 'Sources'
       path_format: 'camel/camel-kafka-connector/{version}/apache-camel-kafka-connector-{version}-src.zip'
       filename_format: 'apache-camel-kafka-connector-{version}-src.zip'
+    - name: 'Binaries'
+      title: 'Connectors download list'
+      link: '/camel-kafka-connector/latest/connectors.html'
 camel-quarkus:
   id: 'camel-quarkus'
   name: 'Apache Camel Quarkus'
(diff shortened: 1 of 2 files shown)


zregvart:
@oscerd and @valdar can you have a look, we would have an [additional row](https://deploy-preview-512--camel.netlify.app/download/#camel-kafka-connector) on the download page for CKC, but also (the partial is shared) on the [release page](https://deploy-preview-512--camel.netlify.app/releases/ckc-0.6.1/#camel-kafka-connector). The drawback is that for all releases it'll go to the [same URL](https://github.com/apache/camel-website/pull/512/files#diff-eed89ae3f61345dc0b09d6d0a6d45c000303f14400590b769d39859834e702aaR92).

oscerd (approved):

oscerd:
Thanks for this, the drawback is not a problem i believe

PR #522, opened 2020-12-27
Rework previews

zregvart:
In order to comply with the ASF policy we need stop using the netlify preview GitHub action, this adds the netlify-cli which can be invoked via `yarn netlify` and changes the workflow to publish previews using it and github-script to add comment on the pull request.


zregvart:
No way to test this other than merging to master, will merge and fix/backout if it fails to work.

PR #550, opened 2021-02-27
2 tiny adjustments

bvahdat:


layouts/partials/releases/camel.html
```diff
@@ -30,6 +30,7 @@ <h2 id="maven"><a class="anchor" href="#maven"></a>Getting the Binaries using Ma
 <p>To use this release in a Spring Boot application, use the <code>camel-spring-boot-dependencies</code> Bill of Materials (BOM):
 <aside>
   <p><mark>Note:</mark> for Camel 3.0.0 and newer releases the <code>groupId</code> for Camel Spring Boot support changed from <code>org.apache.camel</code> to <code>org.apache.camel.springbooot</code>, so be sure to check that the appropriate <code>groupId</code> is used for the Camel version.</p>
+  <p><mark>Note:</mark> for Camel 3.8.0 and newer releases there is another BOM available with the <code>artifactId</code> of <code>camel-spring-boot-bom</code> which you could instead make use of. For details see <a href="/camel-spring-boot/latest/#_camel_spring_boot_bom_vs_camel_spring_boot_dependencies_bom">the corresponding documentation.</a></p>
```
oscerd, line 33: This should be autogenerated somewhere

bvahdat: Not sure if I get your point. There is also another note about `3.0.0` which gets displayed the same way even for the releases before `3.0.0`:

https://camel.apache.org/releases/release-2.25.2/

oscerd: Yes, I meant to say the page is autogenerated, but it is just a one-shot generation, so it can be modified.

PR #564, opened 2021-04-13
refactor: keep the height fixed in Kamelet catalog

zregvart:
I've noticed that layout for some Kamelets in the catalog breaks through
the container, so this reverts the fixed confines to the height instead
of the width. This also has some interesting side-effects (e.g. the
Minio Kamelet), but overall doesn't look as broken as the fixed width.

### DIFF
--- a/antora-ui-camel/src/css/catalog.css
+++ b/antora-ui-camel/src/css/catalog.css
@@ -43,5 +43,7 @@
 }
 
 .catalog .item .image img {
-  width: 4rem;
+  height: 3rem;
+  width: auto;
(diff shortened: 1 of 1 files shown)


zregvart:
cc @nicolaferraro

oscerd (approved):

davsclaus (approved):

nicolaferraro (approved):

oscerd (approved):

zregvart:
Absolute links to camel.apache.org issues were recently fixed in main camel repository, so this is okay to merge.

PR #565, opened 2021-04-13
refactor: use dashed underline link decoration

zregvart:
The current implementation doesn't look as appealing to me, I think the
(almost) default decoration for links looks better.


oscerd (approved):

davsclaus:
Yes this does not look good, I think we should leave it as it is today, people have no problems with finding and clicking the links.

PR #579, opened 2021-05-18
feat(blog): add Twitter share button to blog post

tadayosi:
Let's add Twitter share button to each blog post, so that we can share great articles more easily ;-)

### DIFF
--- a/antora-ui-camel/src/css/sharingbuttons.css
+++ b/antora-ui-camel/src/css/sharingbuttons.css
@@ -1,0 +1,89 @@
+.resp-sharing-button__link,
+.resp-sharing-button__icon {
+  display: inline-block;
+}
+
+.resp-sharing-button__link {
+  text-decoration: none;
+  color: #fff;
+  margin: 0.5em;
+}
+
+.resp-sharing-button {
+  border-radius: 5px;
+  transition: 25ms ease-out;
+  padding: 0.5em 0.75em;
+  font-family: Helvetica Neue, Helvetica, Arial, sans-serif;
+}
+
+.resp-sharing-button__icon svg {
+  width: 1em;
+  height: 1em;
+  margin-right: 0.4em;
+  vertical-align: top;
+}
+
+.resp-sharing-button--small svg {
+  margin: 0;
+  vertical-align: middle;
+}
+
+/* Non solid icons get a stroke */
+
+.resp-sharing-button__icon {
+  stroke: #fff;
+  fill: none;
+}
+
+/* Solid icons get a fill */
+
+.resp-sharing-button__icon--solid,
+.resp-sharing-button__icon--solidcircle {
+  fill: #fff;
+  stroke: none;
+}
+
+.resp-sharing-button--twitter {
+  background-color: #55acee;
+  border-color: #55acee;
+}
+
+.resp-sharing-button--twitter:hover {
+  background-color: #2795e9;
+}
+
+.resp-sharing-button--facebook {
+  background-color: #3b5998;
+  border-color: #3b5998;
+}
+
+.resp-sharing-button--facebook:hover {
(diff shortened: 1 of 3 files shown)


oscerd (approved):

layouts/blog/post.html
```diff
@@ -20,6 +20,13 @@ <h1>{{ .Title }}</h1>
             </figure>
             {{ end }}
             {{ end }}
+            <a href="https://twitter.com/intent/tweet" class="twitter-share-button"
+                data-size="large"
+                data-text="{{ .Title }}"
+                data-url="{{ .Permalink }}"
+                data-hashtags="ApacheCamel"
+                data-show-count="false">Tweet this blog</a>
+            <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
zregvart, line 29: Unfortunately this won't work with our content security policy. I'm also concerned about tracking and adding weight to the page (5 requests, ~400KB). Perhaps we could do https://sharingbuttons.io/ instead?

tadayosi: Thanks Zoran. I was thinking there might be such a case. sharingbuttons.io looks great. So, according to our policy:
https://camel.apache.org/privacy-policy/
we should no embed a js script that tracks users' activities, and as sharingbuttons.io claims that it only uses HTML & CSS and does not track users' activities we can use the widget in the blog template, right?

tadayosi:
OK, this time I used sharingbuttons.io to add three buttons (FB/Twitter/LinkedIn, my opinionated choice). I also added "Share this blog" label before the buttons as otherwise I noticed it may wrongly signal they look SNS links for the author.

layouts/blog/post.html
```diff
@@ -20,6 +20,33 @@ <h1>{{ .Title }}</h1>
             </figure>
             {{ end }}
             {{ end }}
+
+            <h4>Share this blog</h4>
+
+            <!-- Sharingbutton Facebook -->
+            <a class="resp-sharing-button__link" href="https://facebook.com/sharer/sharer.php?u={{ .Permalink }}" target="_blank" rel="noopener" aria-label="">
```
zregvart, line 27: I need to test this, but I think we could use the `aria-label` here to help folk with sight impairment. Same goes for other buttons.

```suggestion
            <a class="resp-sharing-button__link" href="https://facebook.com/sharer/sharer.php?u={{ .Permalink }}" target="_blank" rel="noopener" aria-label="Share on Facebook">
```

antora-ui-camel/src/css/sharingbuttons.css
```diff
@@ -0,0 +1,162 @@
...
+.resp-sharing-button--twitter {
+  background-color: #55acee;
+  border-color: #55acee;
+}
+
+.resp-sharing-button--twitter:hover {
+  background-color: #2795e9;
+}
+
+.resp-sharing-button--pinterest {
```
zregvart, line 55: There are bits that we don't use here (pinterest being one of them), it's not a real issue, couple of bytes more in the CSS, but perhaps we can remove those as we're not using them?

zregvart (commented):
Looks good, thanks, let's wait for the checks and preview.

antora-ui-camel/src/css/sharingbuttons.css
tadayosi: I kept them as-is thinking others might want to add Pinterest share button (or etc.) later :smiley: but I'm fine with trimming unused ones.

layouts/blog/post.html
tadayosi: Agreed. I'll add them.

tadayosi:
Looking good, isn't it?  Now I'll apply Zoran's feedback and update it before merge.
Screenshot from 2021-05-19 17-26-51

PR #586, opened 2021-05-31
chore: redesign the release downloads

zregvart:
With this now we have `eol` front matter that'll be added to the new
_Notes_ column in the download table which, replaces the _Description_
column; also features a link to the release notes, previously linked
from the version number. The description of the download is now in
parentheses next to the download link.

The kind of the download (LTS/latest/legacy) is now designed as a label
to make it more prominent.

Also shortens the column headings for better layout on smaller screens
and compacts the PGP and SHA512 columns into one.

### DIFF
--- a/antora-ui-camel/src/css/download.css
+++ b/antora-ui-camel/src/css/download.css
@@ -6,6 +6,20 @@
   white-space: nowrap;
 }
 
-.download tr td:first-child small {
-  font-style: italic;
+.download tr td:first-child strong {
+  font-size: 0.9rem;
+}
+
+.download tr td em.kind {
+  color: var(--body-background);
+  background: var(--color-asf-light-blue);
+  border-radius: 3px;
+  font-style: normal;
+  font-size: 0.75rem;
+  font-weight: bold;
+  padding: 0.1rem 0.3rem;
+}
+
+.download .tableblock tbody tr:nth-child(2n) {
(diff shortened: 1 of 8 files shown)


zregvart:
Discussion on issue [CAMEL-16669](https://issues.apache.org/jira/browse/CAMEL-16669)

davsclaus (approved):

oscerd (approved):

zregvart:
I've fixed EOL date for 3.4.5, I added EOL dates just for 3.7.4 and 3.4.5. I'm not sure if we should add EOL dates or other indication to the latest releases, say for 3.10.0, or for the legacy releases of 2.x.

@gzurowski you also might want to take note, this adds a `eol` parameter to the front matter, computed by default to be +1 year.

Also cc @ppalaga and @valdar and @nicolaferraro for subproject releases, you might want to chime in, or take advantage of the `eol` parameter.

gzurowski (approved):
LGTM 👍

zregvart:
It's a bit weird that the rows in the table are stripped on the release page and not on the download page, also the release page links to itself which is a bit odd.

Example:
https://pr-586--camel.netlify.app/releases/k-1.4.0/ vs https://pr-586--camel.netlify.app/download/#camel-k

I'll clean that up a bit.

PR #625, opened 2021-09-15
restore canonical link element

djencks:
This restores the <link rel="canonical"... /> element recently removed.


oscerd (approved):

zregvart:
Thanks, I'll follow up with the same change in the Hugo bits.

PR #633, opened 2021-09-24
Remove outdated sub project versions in the documentation

davsclaus:


antora-playbook.yml
```diff
@@ -35,7 +35,7 @@ content:
     - url: https://github.com/apache/camel-k.git
       branches:
         - main
-        - release-1.4.x
+        - release-1.6.x
```
zregvart, line 38: We can't build two branches with the same version, `latest` in this case is the version both in `main` and in `release-1.6.x`. This needs to be set to `1.6.x`:

https://github.com/apache/camel-k/blob/f23eecc4f57d78348abb8b5ba543278f000d2dcd/docs/antora.yml#L3

davsclaus: Thanks Zoran, I have fixed this in camel-k

davsclaus:
Is there a bot command to trigger a re-build ?

zregvart:
> Is there a bot command to trigger a re-build ?

I don't think so, we don't have the trigger configured for that. The Jenkins pipeline can be retried so I've [done that](https://ci-builds.apache.org/blue/organizations/jenkins/Camel%2FCamel.website/detail/PR-633/6/pipeline/), and the GitHub workflow as well.

PR #663, opened 2021-11-05
chore: dependency checks and deduplication

zregvart:
In addition to the cache check which is now run with `--immutable-cache` this also adds a `dedup` check so we don't accumulate duplicate dependencies in the cache.

Builds upon #662


oscerd (approved):

zregvart:
The build is failing due to broken `xref`s in camel-kafka-connector. The checks for dependencies pass, so I'm going to merge this.

PR #669, opened 2021-11-13
resolves #666: finish up with spring-boot changes to use aggregate-collector

### LINKED ISSUE #666: Use aggregate-collector to avoid symlinks; generate pages from json files

djencks:
This brings in 

apache/camel-spring-boot#405
apache/camel-spring-boot#406
apache/camel-spring-boot#407
apache/camel-spring-boot#408
apache/camel-spring-boot#409

Most likely we'll use https://github.com/apache/camel-website/pull/671 instead of this.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -71,20 +71,16 @@ content:
         - docs
         - connectors
 
-    - url: https://github.com/apache/camel-spring-boot.git
-#    - url: https://github.com/djencks/camel-spring-boot.git
+#    - url: https://github.com/apache/camel-spring-boot.git
+    - url: https://github.com/djencks/camel-spring-boot.git
       branches:
-        - main
-        - camel-spring-boot-3.12.x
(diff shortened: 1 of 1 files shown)


djencks:
Closing in favor of #671

PR #680, opened 2021-11-19
Add Camel K 1.7.x and runtime 1.10.x documentation

astefanutti:


zregvart:
It's a bit difficult to parse the error from the build, I think this is the issue:
```
Duplicate nav in next@camel-k: modules/ROOT/nav-end.adoc
```

astefanutti:
> It's a bit difficult to parse the error from the build, I think this is the issue:
> 
> ```
> Duplicate nav in next@camel-k: modules/ROOT/nav-end.adoc
> ```

Thanks @zregvart. My understanding is that comes from https://github.com/apache/camel-k/pull/2757 that has to be merged first. But I prefer having confirmation from the experts :)

Issue #686, opened 2021-11-24
Add more prominent link to the official Apache Camel group on linkedin


davsclaus:
The official group
https://www.linkedin.com/groups/2447439/

We should add a link to this group in a more prominent place.
Maybe next to twitter icon in the top right corner, if we can find a icon in orange color scheme

PR #695, opened 2021-11-29
#686: Add icon to Apache Camel group on Linkedin

davsclaus:


oscerd (approved):

antora-ui-camel/src/img/brand-logos.svg
```diff
@@ -20,4 +20,9 @@
     <desc>Chat on Zulip</desc>
     <path d="m 131.6034,4 5.79516,12.766113 h 23.09009 L 131.44616,46.136718 137.94714,60 h 41.74537 L 173.0343,46.136718 H 150.95764 L 179.22253,16.608887 173.11291,4 Z" />
   </symbol>
+  <symbol id="linkedin" viewBox="192 0 64 64">
+    <title>Linkedin</title>
+    <desc>Apache Camel group on Linkedin</desc>
+    <path d="m 131.6034,4 5.79516,12.766113 h 23.09009 L 131.44616,46.136718 137.94714,60 h 41.74537 L 173.0343,46.136718 H 150.95764 L 179.22253,16.608887 173.11291,4 Z" />
```
zregvart, line 26: This seems to be copy-pasted icon of Zulip, there are icons available from [LinkedIn](https://brand.linkedin.com/downloads), though they seem to be in PNG and we want SVG icons here. We should also review that we comply with their [policies](https://brand.linkedin.com/policies).

zregvart (changes requested):
Not shown on my end, probably not the best idea to use the Zulip icon.
Screenshot from 2021-11-29 12-28-56

antora-ui-camel/src/img/brand-logos.svg
davsclaus: Yes its a copy/paste of that icon, to ensure that a icon is shown. As I could not get it to show the linkedin icon.

davsclaus: Since the icon is not shown on your end too, it may be something wrong with the preview mode, and that its shown on the real website.

zregvart: Preview is correct. The copy pasted Zulip icon has geometry that is outside of the viewbox, even if it was shown it would be confusing to the users to see two icons for Zulip.

davsclaus: Okay so the icon works - and we can replace it with an actual linkedin icon.

davsclaus:
Icon is shown now
Screenshot 2021-11-30 at 09 15 30

PR #700, opened 2021-11-29
Resolves #699 hugo docs page directs to latest not next docs

### LINKED ISSUE #699: /docs/ page should direct to latest, not next, versions of docs
Directing users to the 'next' documentation implies that the unreleased 'next' version of code can be used for purposes other than development of camel, against Apache policy.

djencks:
See #699.
The netlify preview won't show this working as it relies on the .htaccess file.  I checked that it worked locally with httpd running in docker.

### DIFF
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -5,7 +5,7 @@ Title: "Documentation"
 
 {{< div "icon" >}}
 
-[![Camel](/_/img/logo-d.svg)](/camel-k/next/)
+[![Camel](/_/img/logo-d.svg)](/camel-k/latest/)
 
 {{< /div >}}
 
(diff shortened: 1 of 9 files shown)


oscerd (approved):

zregvart:
Not sure why are we doing this? Didn't we agree that to replace `"latest"` with `"next"`? Why link to `"latest"` at all?

djencks:
See #699 :-)  Directing people to the next prerelease docs IMO violates apache policy that unreleased code is only for development purposes.

I'm very unhappy with how I'm doing this but can't think of  a better way.  If Antora was building the docs page it would be easy, but I can't think of a way of supplying the actual version of the latest docs to hugo, so I'm sending to `latest` which redirects  to the actual latest version.  This is causing the link checker to object, so I'm adding actual static pages at the link targets that should be concealed by the redirect rules.

PR #705, opened 2021-12-07
chore: run redirect test as a part of checks

zregvart:
This makes sure that we run the redirect tests as a part of regular
checks so we don't unintentionally break redirects.

The `test/redirect.sh` now supports these new parameters:
 - `-s` - start a container with Apache HTTP server and test against it
 - `-v` - verbose test output (print a line for each test)
 - `-x` - verbose script output (equivalent of `set -x`)


oscerd (approved):

zregvart:
Seems to be stable now. I'll add a note on this in a README, but I really want to merge this now.

PR #712, opened 2021-12-09
feat: cache bursting for UI images

### LINKED ISSUE #702: Use cache bursting hash on resources from the Antora UI
Recently we updated the `antora-ui-camel/src/img/brand-logos.svg`, but the effect of that change might not be seen by all user agents depending on their cache state. The cache expires [after 1 month](https://github.com/apache/camel-website/blob/edf4924352b466056ad6bc8f00744dd199d0dfeb/static/.htaccess#L1901), making the user agent skip fetching the resource until it expires, which could be a ~30 days from now.
For CSS we use cache bursting hash, we should use it for other resources within the Anotra UI so changes are immediately in effect.

zregvart:
feat: cache bursting for UI images

This configures the UI build to generate files with the hash within them for images in addition to the CSS and JavaScript files we had. Also layouts/partials that were pointing to the un-hashed versions were updated.

Since images are quite possibly linked from other websites, most likely the logo image, a set of redirect rules was added not to break those sites.

Fixes #702


zregvart:
I think this is now ready to merge, I'll leave it for a bit as it is a larger change, to give time for comments.

PR #713, opened 2021-12-10
fix: pass only the PR number and validate it

zregvart:
We don't need to pass `github.event.pull_request.html_url` in
`./pull_request/html_url` in the archive, and we should validate that
`./pull_request/number` is a number and we should read only the first
line from that file.


oscerd (approved):

zregvart:
I'm going to merge this, any feedback and improvements on this can be dealt with in a new pull request.

PR #719, opened 2021-12-14
fix: generate the assets.js helper

### LINKED ISSUE #718: Some pages are showing as 'white background' only
https://camel.apache.org/manual/getting-started.html
https://camel.apache.org/components/3.13.x/index.html
https://camel.apache.org/manual/writing-components.html

are all returning proper HTML but appear as white background...

Tried from multiple browsers (Chrome, Firefox, Edge)
and from multiple machines

other pages seem to look ok (homepage included)

zregvart:
With #702 the Gulp pipeline changed behavior, previously invoking
`next(null, null)` was generating the `assets.js` helper now that is no
longer sufficient. By design of `through2` invoking the callback with
`null, null` doesn't emit the file, so not sure why that was working
previously.

Also since adding `rev-rewrite` to the pipeline we no longer need to use
the `asset` helper for non-dynamic resource usage, i.e. those that
specify the file name, as those will be replaced by `rev-rewrite` for
their hashed counterparts.

Fixes apache#718

### DIFF
--- a/antora-ui-camel/gulp.d/tasks/build.js
+++ b/antora-ui-camel/gulp.d/tasks/build.js
@@ -134,7 +134,7 @@ module.exports = (src, dest, preview) => () => {
     .pipe(
       map((file, enc, next) => {
         manifest = file.contents.toString()
-        next(null, null)
+        next(null, file)
       })
     )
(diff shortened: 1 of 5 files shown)


zregvart:
Preview looks okay, merging.

PR #729, opened 2021-12-20
DO NOT MERGE Main content as tables preview

djencks:
This is a PR to build a preview of the work around https://github.com/apache/camel-quarkus/issues/3396. It also illustrates that it's now pretty easy to construct a preview PR for multiple doc branches (10 in this case).

DO NOT MERGE.  This is only to build a preview.


djencks:
The changes this previews are applied.

PR #738, opened 2021-12-24
Resolves #732: Main commit UI bundle

### LINKED ISSUE #732: Commit the built Antora ui bundle so it's always available
The local/partial build setup needs the Antora UI bundle to be available.

Building it locally is at best a nuisance and on MacOS requires running in Docker or running yarn install to set up native dependencies for MacOS rather than the checked-in linux versions.

Another solution was proposed in https://github.com/apache/camel-website/pull/537, along with some variants, but I believe publishing the UI bundle as proposed there is significantly more complex for every use of the bundle with no redeeming advantages.

I'm happy with just manually committing it whenever it changes and we notice, since the official builds and previews will continue to build it just before using it, but it would be even better for the PR build or jenkins build to open a PR when it notices that the UI bundle has changed.

djencks:
This:
- commits the UI bundle
- Adds a GitHub actions workflow adapted from the main camel `master-push-build.yml` that is supposed to build the UI bundle and check for differences and create a PR.  THIS NEEDS REIEW (IMHO).  In particular I left the `assignees: oscerd` because I don't know the syntax.  I'm happy to review, perhaps Zoran is also.
- Adds a `yarn-in-docker.sh` script that lets one (me) run yarn commands that don't work on MacOS.


oscerd (approved):
Any GH account could be an assignee, i think we have an example on camel-kamelets with 2 assignees

djencks:
This needs some more work, it turns out that the ui bundle build is not reproducible, so right now this would generate a PR on every push whether or not it changed the UI.

djencks:
Fixed: we can now use the unpacked UI bundle in Antora, and that part is reproducible.
I added myself to the list of auto-assignees.  Zoran, can I add you too?

djencks:
Zoran, I merged this and would still appreciate a review when you have a chance.

zregvart:
Just refined this a bit in #750, I think we should also document some of this, perhaps in the README -- I can update that

Issue #733, opened 2021-12-21
Move the published website to a separate repo


djencks:
The local/partial builds idea requires everyone working on the docs (hopefully all camel devs :-) to clone and pull the camel-website repo moderately often.  The main branch is pretty tiny compared with  the published site, and although it's possible to just clone a single branch it's rather tricky.

It's confusing to pull and see a big update as the published site is completely replaced when there is no change in main.

It's a pretty simple change to publish to a separate repo rather than a branch in the current repo.

Publishing to a separate repo offers the ability to keep website history rather than only the current version (by squashing commits) without slowing down everything in sight, but whether to do that is a separate question.

djencks:
Additional steps needed (I think):

- create `camel-website-pub` repo (I think I can't do this)
- copy the asf-site branch to it, and make sure the initial commit still has the same oid.

PR #741, opened 2021-12-27
Resolves issue #733: modifications to publish to separate git repo

djencks:
I think these "code" changes are what's needed to publish to a separate repo `camel-website-pub.git`.

See issue for additional steps.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -94,7 +94,7 @@ pipeline {
             steps {
                 dir('deploy/live') {
                     deleteDir()
-                    sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website.git .'
+                    sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website-pub.git .'
                     sh "git -c core.editor='sed -i 2,/\$(git log --skip=9 -1 --pretty=format:%h)/s/^pick/squash/' rebase -q --interactive $STOP_SQUASH_AT" // squash all but initial and last 9 commits
                     sh 'git rm -q -r *'
                     sh "cp -R $WORKSPACE/camel-website/public/. ."
--- a/static/.asf.yaml
+++ b/static/.asf.yaml
@@ -18,6 +18,8 @@
 notifications:
   pullrequests: commits@camel.apache.org
   jira_options: link label comment
+publish:
+  whoami: asf-site
 github:
   labels:
     - camel


.asf.yaml
```diff
@@ -17,18 +17,6 @@
 
 notifications:
   pullrequests: commits@camel.apache.org
-  jira_options: link label comment
```
zregvart, line 20: Looks like a typo

.asf.yaml
```diff
@@ -17,18 +17,6 @@
...
+  jira_options: link label=
 publish:
   whoami: asf-site
```
zregvart, line 22: This needs to be removed if we're not publishing from the `asf-site` branch

.asf.yaml
```diff
@@ -17,18 +17,6 @@
...
-    - camel
-    - integration
-    - java
-    - hacktoberfest
-  enabled_merge_buttons:
-    squash:  true
-    merge:   false
-    rebase:  true
-  features:
-    issues: true
```
zregvart, line 34: This has little to do with publishing the website, accidentally removed?

Jenkinsfile
```diff
@@ -94,7 +94,7 @@ pipeline {
             steps {
                 dir('deploy/live') {
                     deleteDir()
-                    sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website.git .'
+                    sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website-pub.git .'
                     sh "git -c core.editor='sed -i 2,/\$(git log --skip=9 -1 --pretty=format:%h)/s/^pick/squash/' rebase -q --interactive $STOP_SQUASH_AT" // squash all but initial and last 9 commits
```
zregvart, line 98: `STOP_SQUASH_AT` needs to be updated

static/.asf.yaml
```diff
@@ -0,0 +1,34 @@
+#
```
zregvart, line 1: We should keep this file in `camel-website-pub` repository

zregvart (changes requested):
The new repository can be created [here](https://gitbox.apache.org/setup/newrepo.html) (any PMC member can do that). This needs more work.

.asf.yaml
djencks: agreed.

.asf.yaml
djencks: Well, I agree it should be, but if there's no longer an asf-site branch in this repo it's just confusing rather than an error:-)

.asf.yaml
djencks: yes.

Jenkinsfile
djencks: I don't think we can do that until the asf-site branch is moved to the new repo and we determine if the OID is the same.... I think it should be, but I'm not sure.

static/.asf.yaml
djencks: On this I disagree with you.  If we keep it here, publishing consists of removing _everything_ from the site-pub repo, adding the entire site, and committing.  If we keep it in  the `camel-website-pub` repo then we have to make an exception for one file.  I think that is excessively complex for no gain.

djencks:
I believe I'm no longer a PMC member so someone else will have to create the repository, if we proceed with this.

Jenkinsfile
djencks: I also think we should stop squashing, and just keep all the history.  It doesn't slow down our website deploy, in fact it should be marginally quicker as no squash is needed.  If we tag the site just before we remove an outdated subproject version, we can point then people to the old documentation rather easily.

.asf.yaml
zregvart: This is read by the tools used to publish the website, see [chapter 5. Publishing](https://infra.apache.org/project-site.html), and might end up causing issues when publishing.

static/.asf.yaml
zregvart: As far as I understand the [documentation](https://infra.apache.org/project-site.html) on publishing the website, the built website from camel-website-pub repository will not be published without it. This is a requirement for publishing the website that the git repository with the website content contains `.asf.yaml` file.

zregvart:
> I believe I'm no longer a PMC member so someone else will have to create the repository, if we proceed with this.

I've created the repository (https://gitbox.apache.org/repos/asf?p=camel-website-pub.git, https://github.com/apache/camel-website-pub)

static/.asf.yaml
djencks: If we keep it in static/.asf.yaml, it will be placed in the correct branch in camel-website-pub with every commit. I set up the Aries and Felix websites this way and they seem to work fine.

djencks:
Thanks for creating the repo!  How shall we proceed?

I think we need to:
- create an asf-site branch in the new repo with at least one commit. I guess we can push the asf-site branch from camel-website to the new repo.
- merge this PR
- see what happens with  the Jenkins build.

I'm happy to do this or have you do it.  My guess is that you know more about backing stuff out if there are problems.

static/.asf.yaml
zregvart: Oh, that's actually quite clever, I missed that, apologies :+1:

zregvart:
> Thanks for creating the repo! How shall we proceed?
> 
> I think we need to:
> 
>     * create an asf-site branch in the new repo with at least one commit. I guess we can push the asf-site branch from camel-website to the new repo.
> 
>     * merge this PR
> 
>     * see what happens with  the Jenkins build.

I think we can try something like that, perhaps merge this first. I wonder what will happen if we have two asf-site branches, I guess if there would be any disruption it won't take too long.

Nonetheless, I'll put out an email and notify folk on Zulip so they aren't surprised if the website is unavailable for a bit.

I'll keep a copy of everything so we can backtrack if needed.

PR #763, opened 2022-01-23
Main 701 ri tables

djencks:
This must be merged in coordination with several other PRs.

This sets up some global helper functions for the RI tables and adds the new branches for camel-k 1.8.x,camel-k-runtime 1.11.x, and kamelets 0.7.x releases. As of this PR creation, the vote has gone on for more than 72 hours but has not concluded.

### DIFF
--- a/util/jsonpath-util.js
+++ b/util/jsonpath-util.js
@@ -105,6 +105,41 @@ module.exports = {
   valueAsString: (value) => {
     return value === undefined ? '' : `${value}`
   },
+
+//  Compatibility table support
+  camelRef: (version, docVersion) => (docVersion === 'none') ? version : `xref:${docVersion}@components:ROOT:index.adoc[${version}]`,
+
+  ckRef: (version, docVersion) => (docVersion === 'none') ? version : `xref:${docVersion}@camel-k:ROOT:index.adoc[${version}]`,
+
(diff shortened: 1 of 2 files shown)


djencks:
This will fill in all those new RI tables.

util/jsonpath-util.js
```diff
@@ -105,6 +105,41 @@ module.exports = {
...
+    const v2 = t2['page-component-version']
+    if (v1 === v2) return 0
+    if ((v1 === 'next') || !v2) return -1
+    if ((v2 === 'next') || !v1) return 1
+    const v1s = v1.split('.').map((t) => Number(t))
+    const v2s = v2.split('.').map((t) => Number(t))
+    let i
+    for (i = 0; i < v1s.length; i++) {
+      if (i === v2s.length) return -1
+      if (v1s[i] !== v2s[i]) return v2s[i] - v1s[i] // could be a problem with NaN?
```
zregvart, line 139: If we stick with equal (3) numbers in each version, I don't think so. If this becomes a bit more complicated we could use [semver](https://www.npmjs.com/package/semver), no need now I think, this is straightforward.

zregvart (approved):

davsclaus (approved):

PR #765, opened 2022-01-24
feat: add 'Supports Java 11' for Kamelets

zregvart:
Given that Kamelets have an Catalog API we should inform folk that Java
11 is required.


oscerd (approved):

davsclaus:
Yes it works now Java 11 is shown

PR #767, opened 2022-01-26
fix: social link dynamically retrieved

### LINKED ISSUE #679: Some problem with social media overview
I've just published a tweet announcing the last blog published: https://twitter.com/Squ4k3/status/1461290119575621636

However the overview does not point to the description, and, in fact, the `meta` is falling back to the main page description instead:
```
<meta property="og:description" content="Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.">
```
What is the proper way to handle the specific description?

squakez:
Closes #679

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -17,12 +17,25 @@
     <link rel="icon" type="image/png" href="{{ "favicon-16x16.png" | relURL }}" sizes="16x16">
     <link rel="icon" type="image/png" href="{{ "favicon-128.png" | relURL }}" sizes="128x128">
     <meta name="application-name" content="Apache Camel">
-    <meta property="og:title" content="Apache Camel helps you integrate over 300 different systems!">
-    <meta property="og:site_name" content="Apache Camel">
-    <meta property="og:url" content="https://camel.apache.org">
-    <meta property="og:description" content="Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.">
+    
+    <meta property="og:title" content="{{ .Title }}"> 
+    <meta property="og:site_name" content="{{ .Site.Title }}"> 
+    <meta property="og:url" content="{{ .Permalink }}"> 
+    {{ if .Description}}
+        <meta name="og:description" content="{{ .Description }}">
+    {{ else }}
+        {{ if .Params.preview}}
+        <meta name="og:description" content="{{ .Params.preview }}">
+        {{ end }}
+    {{ end }}
     <meta property="og:type" content="website">
-    <meta property="og:image" content="{{ .Site.Params.organizationLogo }}">
+    {{ $featured := (.Resources.ByType "image").GetMatch "*featured*" }}
+    {{ with $featured }}
+        <meta property="og:image" content="{{ $featured.RelPermalink }}">
+    {{ else }}
+        <meta property="og:image" content="{{ .Site.Params.organizationLogo }}">
+    {{ end }}
+    
     <link rel="manifest" href="{{ "site.webmanifest" | relURL }}">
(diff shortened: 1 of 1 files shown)


squakez:
Ref https://issues.apache.org/jira/browse/CAMEL-16998

layouts/partials/header.html
```diff
@@ -17,12 +17,25 @@
...
-    <meta property="og:title" content="Apache Camel helps you integrate over 300 different systems!">
-    <meta property="og:site_name" content="Apache Camel">
-    <meta property="og:url" content="https://camel.apache.org">
-    <meta property="og:description" content="Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.">
+    
+    <meta property="og:title" content="{{ .Title }}"> 
+    <meta property="og:site_name" content="{{ .Site.Title }}"> 
+    <meta property="og:url" content="{{ .Permalink }}"> 
+    {{ if .Description}}
+        <meta name="og:description" content="{{ .Description }}">
```
zregvart, line 25: Where does the `.Description` come from? I can see only that some security announcements have a `description` front matter set.

layouts/partials/header.html
```diff
@@ -17,12 +17,25 @@
...
+    {{ else }}
+        {{ if .Params.preview}}
+        <meta name="og:description" content="{{ .Params.preview }}">
+        {{ end }}
+    {{ end }}
     <meta property="og:type" content="website">
-    <meta property="og:image" content="{{ .Site.Params.organizationLogo }}">
+    {{ $featured := (.Resources.ByType "image").GetMatch "*featured*" }}
+    {{ with $featured }}
+        <meta property="og:image" content="{{ $featured.RelPermalink }}">
```
zregvart, line 34: Perhaps we should also have in mind the [best practices](https://developers.facebook.com/docs/sharing/best-practices#images) for Open Graph images, and think if we should resize the image or place the width/height properties as well.

layouts/partials/header.html
```diff
@@ -17,12 +17,25 @@
...
+    
+    <meta property="og:title" content="{{ .Title }}"> 
+    <meta property="og:site_name" content="{{ .Site.Title }}"> 
+    <meta property="og:url" content="{{ .Permalink }}"> 
+    {{ if .Description}}
+        <meta name="og:description" content="{{ .Description }}">
+    {{ else }}
+        {{ if .Params.preview}}
+        <meta name="og:description" content="{{ .Params.preview }}">
+        {{ end }}
```
zregvart, line 29: I think we're missing an `{{ else }}` here, if there is no `page.Description` and no `page.Params.preview` defined.

zregvart (commented):
Thanks, let's have a look at the preview. I think the issue is a bit broader than the change proposed here, so I just noted a few outstanding concerns. We can approach this incrementally and start with this, I would ask in that case that issues are created for following up on this.

layouts/partials/header.html
squakez: No idea. I am reusing the same definition for meta-description: https://github.com/apache/camel-website/pull/767/files#diff-42986cf40ad96f5b7ca39d9b71ba6a9f252656de24cfe82100892ff0f4400f90L29 - if it is presents, then it should be used, otherwise we may fallback to other variables, such as the preview that we use in the blog.

layouts/partials/header.html
squakez: I agree, though I'd propose a follow-up issue to fix it properly.

squakez:
Right now the result would be like this:
image
We must work on improving the image (which is required to be 1080p to render automatically). I will take care to create proper follow up issues and work toward resolution. Wdyt @zregvart ?

PR #787, opened 2022-02-14
Added a blog post about test infra and test component deprecations

orpiske:


oscerd (approved):

davsclaus (approved):

zregvart:
> For some reason the post is not appearing on the preview.

Hmm, it should we have `--buildFuture` here:

https://github.com/apache/camel-website/blob/0437c84708e9ea7d716ed6bdca9f74b354a6b069/.github/workflows/pr.yaml#L24

orpiske:
> > For some reason the post is not appearing on the preview.
> 
> 
> 
> Hmm, it should we have `--buildFuture` here:
> 
> 
> 
> https://github.com/apache/camel-website/blob/0437c84708e9ea7d716ed6bdca9f74b354a6b069/.github/workflows/pr.yaml#L24

Thanks. I think it was caused by a draft: true/false in the post. After I removed, it seems to be OK.

davsclaus:
The blog post is in the preview now - but the date is tomorrow ;)

orpiske:
> The blog post is in the preview now - but the date is tomorrow ;)

So, I am thinking about publishing it tomorrow, to give some more time for the Aurelien's and Marat's post to remain on top.

PR #792, opened 2022-02-28
fix: prevent scroll adjustment on type

### LINKED ISSUE #790: Camel components doc page is scrolling when user is typing
When user is typing in the searchbar the page scrolls a little bit, it's not horrible but it's kinda funny.
See gif: 
Peek 2022-02-24 15-43

zregvart:
Seems that the input field to search the component tree is just below
the `scroll-padding-top` so when typing in the input field browser tries
to put it in focus and since it's position is absolute it can never be
in focus according to the browser calculation, so the browser keeps
trying to put it in focus by adjusting the `scroll-top`.
This removes the `scroll-padding-top`, to prevent the browser from
scrolling the page while typing in the input field. Given that the
`body` is not scrollable at all this should not have any other effect
than to remove this unwanted behavior.

Fixes #790


oscerd (approved):

zregvart:
Preview looks okay, merging.

PR #801, opened 2022-03-21
feat: use `GITHUB_TOKEN` for API authentication

zregvart:
This makes use of the `GITHUB_TOKEN` environment variable, if defined,
to authenticate the API requests. Since the `GITHUB_TOKEN` is provided
to the GitHub workflow automatically[1], this should help with rate
limits when building the website.

[1] https://docs.github.com/en/actions/security-guides/automatic-token-authentication


oscerd (approved):

zregvart:
Based on #803 (so only this needs to be merged), should help with detecting errors by providing labels next to the script being run and properly handling exit codes of running hugo.

zregvart:
This seems to fix the GitHub API rate limits issue by authenticating with the auto-generated GITHUB_TOKEN. The issue now is that the examples in `camel-kamelets` contain URLs which are then converted to HTML links that point to non-existent files and are failing the link check.

PR #802, opened 2022-03-21
Added blog post about the Resume API v2

orpiske:


zregvart:
Would most likely need to be rebased on #801

orpiske:
> Would most likely need to be rebased on #801

Oh, thanks for the heads up. I'll keep an eye on that one and will rebase.

davsclaus (approved):

PR #803, opened 2022-03-21
chore(deps): upgrade hugo-extended to 0.95.0

zregvart:
Also upgrades Hugo to the same version. Due to a incompatibility of Yarn
PnP with `imports` in `package.json`[1] declared by chalk version 5+,
the resolution of chalk was set to the latest pre-5 version. Otherwise
the postinstall script of hugo-extended would fail.

[1] yarnpkg/berry#3843


zregvart:
This will get merged via #801

PR #805, opened 2022-03-22
Fixed image links in the resume api post

orpiske:


orpiske:
Screencasts previews don't work after publication. We need to link to the image directly. This should fix it.

PR #818, opened 2022-04-07
Blogpost announcement for VS Code 0.1.1 release

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>


apupier:
failure due to https://github.com/apache/camel-website/issues/819

Will need to update the publication date when main branch is fixed

davsclaus:
The check is green, so this should be good to merge

PR #825, opened 2022-04-13
Promote the usage of yarn in docker

essobedo:
## Motivation

Having the ability to launch the `yarn`  commands inside a Docker container is very handy especially if we don't want to install everything or we use Mac OS, so it could be nice to promote it in the Readme

## Modifications

* Describe how to use it in the readme
* Add a file of environment variables to easily set the GitHub token inside the container
* Add bat file for Windows users
* Publish the port 1313 to have access to the preview
* Bind the Hugo server to `0.0.0.0` to have access to the preview from the Docker host

### DIFF
--- a/yarn-in-docker.sh
+++ b/yarn-in-docker.sh
@@ -5,4 +5,4 @@
 # $yarn-in-docker.sh workspace antora-ui-camel run build
 
 docker build -t camel-website .
-docker run --rm -it -v $(pwd):/work:Z --workdir /work camel-website yarn $*
+docker run --rm -it -p 1313:1313 -e "GITHUB_TOKEN=${GITHUB_TOKEN}" -e "CAMEL_ENV=${CAMEL_ENV}" -v $(pwd):/work:Z --workdir /work camel-website yarn $*
--- a/README.md
+++ b/README.md
@@ -373,6 +373,22 @@ This is most likely what you want.
 If you need to link to a specific non-latest version of a page, specify the version in the xref.
 - A link from anywhere to a camel-quarkus 2.0.0 page would be xref:2.0.0@camel-quarkus::reference/extensions/activemq.adoc[].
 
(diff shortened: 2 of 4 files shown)


yarn-in-docker.sh
```diff
@@ -5,4 +5,4 @@
 # $yarn-in-docker.sh workspace antora-ui-camel run build
 
 docker build -t camel-website .
-docker run --rm -it -v $(pwd):/work:Z --workdir /work camel-website yarn $*
+docker run --rm -it --env-file yarn-in-docker.env -v $(pwd):/work:Z --workdir /work camel-website yarn $*
```
zregvart, line 8:
```suggestion
docker run --rm -it -e GITHUB_TOKEN=${GITHUB_TOKEN} -e CAMEL_ENV=${CAMEL_ENV} -v $(pwd):/work:Z --workdir /work camel-website yarn $*
```

zregvart (commented):
Perhaps we don't need the `yarn-in-docker.env` file for passing two environment variables. If it remains we should add it to `.gitignore` so folk don't accidentally include it when proposing changes.

yarn-in-docker.sh
essobedo: It is not really a flexible approach if for some reason we need more env variables, we will then have to modify the script too moreover you still need to set the env variables locally

essobedo: However if you prefer your approach, I can change it. Let me know WYT

zregvart: I'm okay either way. My motivation was simplicity. We most likely won't be adding new environment variables and we'll remove the support for `HUGO_PARAMS_GitHub*` environment variables in the future, so having two `-e` parameters and having those be inherited from the shell seemed simpler to me.

essobedo: As you wish

PR #826, opened 2022-04-13
Fix the name of the playbook in the preview command

essobedo:
## Motivation

When launching the preview we get the next error indicating that the playbook is missing.
```
[preview:antora] [18:17:57.510] FATAL (antora): playbook file not found at /camel-website/antora-playbook.yml (cwd: /camel-website, playbook: antora-playbook.yml)
[preview:antora] Add the --stacktrace option to see the cause of the error.
```

## Modifications:

* Change the target playbook to set the new playbook name


oscerd (approved):

package.json
```diff
@@ -24,7 +24,7 @@
     "update:cache": "yarn workspaces foreach install --check-cache",
     "update:dedupe": "yarn workspaces foreach dedupe",
     "preview": "run-s -l preview:antora preview:hugo",
-    "preview:antora": "antora --clean --fetch antora-playbook.yml",
+    "preview:antora": "antora --clean --fetch antora-playbook-production.yml",
```
zregvart, line 27: Perhaps remove `preview:antora` altogether and replace `preview:antora` in `preview` script with `build:antora` instead.

```suggestion
```

zregvart (commented):
`preview:antora` doesn't really work unless `build:antora-prep` was invoked beforehand. And running `preview:antora` never stood up a preview server anyhow. So perhaps instead we should do away with `preview:antora` and use `build:antora` instead.

PR #847, opened 2022-05-16
blogpost for release of VS Code Debug Adapter for Camel 0.3.0

apupier:


apupier:
argh

```
[check:links    ] Found invalid urls in blog/2022/05/release-vscode-debug-adapter-0.3.0/index.html:
[check:links    ] 	Linked file at path components/latest/languages/simple-language.html does not exist!
```
so `[Camel Simple language documentation page](/components/latest/languages/simple-language.html)` is incorrect

zregvart:
@apupier files under `latest` don't exist so the link checker marks the link as broken. You can either use a fixed version such as `3.16.x` you used or `next` for the pre-release version. Out of those `3.16.x` remains stable until we remove the docs for it, and `next`'s stability varies. I think `3.16.x` is correct in this context, and we'll probably need to bump it once we remove that version. To which the link checker will alert us again.

davsclaus (approved):

davsclaus:
Can we change the date to today before we merge it

oscerd (approved):

apupier:
> Can we change the date to today before we merge it

Updated

> @apupier files under `latest` don't exist so the link checker marks the link as broken. You can either use a fixed version such as `3.16.x` you used or `next` for the pre-release version. Out of those `3.16.x` remains stable until we remove the docs for it, and `next`'s stability varies. I think `3.16.x` is correct in this context, and we'll probably need to bump it once we remove that version. To which the link checker will alert us again.

Thanks for the explanation. I set `3.16.x` then.

PR #866, opened 2022-06-27
Use fewer clicks to get to the sub-project pages

orpiske:
Recently I received a feedback in the community that our documentation takes too many clicks to get to the information. The flow from "Read More" to the actual documentation was one of the examples. This adjusts it to go directly to the sub-project pages.

What do you think?

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -123,7 +123,7 @@ Apache Camel helps you integrate various systems consuming or producing data.
 
 ## Camel K
 Apache Camel K is a lightweight integration framework that runs natively on Kubernetes.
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

zregvart:
@orpiske as the message says that is to be expected when previewing on Netlify; and if the link checks passed (and they did), this should be okay.

orpiske:
> @orpiske as the message says that is to be expected when previewing on Netlify; and if the link checks passed (and they did), this should be okay.

Awesome, thanks for confirming it

PR #907, opened 2022-09-23
feat: switch to using Goldmark

zregvart:
Hugo deprecated Blackfriday Markdown renderer in favor of Goldmark which is the new default. A warning that Blackfriday will be removed in a future version is displayed when rendering the pages. To achieve the same functionality as we had with Blackfriday, as the `rel` attribute for links is not configurable, we need to add `render-link.html` partial.
Other changes include enabling inline HTML
(`markup.goldmark.renderer.unsafe=true`) and disabling inline CSS for syntax higlighting. We did not allow inline CSS on the production site, but in previews are not able to set HTTP headers, so the Content Security Policy header can't be set to disable inline CSS. We can set the option `markup.higlight.noClasses=false` that the inline CSS is not added to the `<code>` blocks and the styling is performed via `higlight.js`. This helps the preview to render syntax higlighted code blocks as they would be rendered on the production site.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -9,10 +9,13 @@ enableRobotsTXT = true
 timeout = 300000
 
 [markup]
-defaultMarkdownHandler = "blackfriday"
-    [markup.blackfriday]
-        nofollowLinks = true
-        noreferrerLinks = true
+defaultMarkdownHandler = "goldmark"
+    [markup.goldmark]
+        autoHeadingIDType = "blackfriday"
+    [markup.goldmark.renderer]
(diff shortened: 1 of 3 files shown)


zregvart:
I created this as draft, I'm not sure if there is anything that I've overlooked. I would appreciate if folk could have a look at the preview and see if they can spot anything different.

oscerd (approved):

apupier (approved):
browsed the preview and did not noticed any troubles

PR #928, opened 2022-11-28
fix: og:image

### LINKED ISSUE #768: Tune image output for social links
Right now the result would be like this:
image
We must work on improving the image (which is required to be 1080p to render automatically).

_Originally posted by @squakez in https://github.com/apache/camel-website/issues/767#issuecomment-1022310649_

squakez:
Closes #768

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -17,10 +17,10 @@
     <link rel="icon" type="image/png" href="{{ "favicon-16x16.png" | relURL }}" sizes="16x16">
     <link rel="icon" type="image/png" href="{{ "favicon-128.png" | relURL }}" sizes="128x128">
     <meta name="application-name" content="Apache Camel">
-    
-    <meta property="og:title" content="{{ .Title }}"> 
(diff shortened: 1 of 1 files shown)


layouts/partials/header.html
```diff
@@ -31,11 +31,11 @@
     <meta property="og:type" content="website">
     {{ $featured := (.Resources.ByType "image").GetMatch "*featured*" }}
     {{ with $featured }}
-        <meta property="og:image" content="{{ $featured.RelPermalink }}">
+        <meta property="og:image" content="https://camel.apache.org{{ $featured.RelPermalink }}">
```
squakez, line 34: @zregvart I finally made some test and your solution looks good. Do you know if there is some property on the `$featured` image to directly get the full url by any chance?

zregvart, line 34: Try with this:

```suggestion
        <meta property="og:image" content="{{ $featured.Permalink }}">
```

squakez:
Cool, I think it's ready to be merged:
image

PR #929, opened 2022-12-01
CAMEL-18786: support better organization for the Camel Core documentation

orpiske:
These changes allow the Camel Core project to create a new page to
reorganize, categorize and separate the project documentation according
to the audience.

The behavior of a few of the static pages is changed to accommodate
that.


orpiske:
This depends on apache/camel#8821.

orpiske:
The preview and more details are available on the linked PR.

orpiske:
@zregvart @djencks hey guys, any idea what might be the problem here? The build is failing with:

```
➤ YN0000: [build:minify  ] [09:56:21] Error in plugin "gulp-htmlmin"
```

I didn't get those errors locally, so I am not sure what might be the problem. 

If you have any pointers, that would be helpful. Thanks!!!

orpiske:
> There is a typo on this line:
> 
> https://github.com/apache/camel/blob/ea35b3ee5c554bb56f7434267661b4ede723df01/docs/main/modules/ROOT/attachments/contents-table.html#L60
> 
> `<bcomponents...` is not a valid HTML tag.
> 
> If you want to run all the checks locally run with `CAMEL_ENV=production` [environment variable](https://github.com/apache/camel-website#camel_env-environment-variable).

Awesome! Thanks for pointing the error and sharing the info about the checks. I'll make sure to include that into my workflow.

> 
> In general I don't think it is a good idea to have HTML outside of the website repository, things are prone to breaking that way. The layouts that generate the HTML need to be in the website directory.

I agree that it would be better to avoid HTML. Unfortunately I couldn't get it to look precisely the way I wanted. 

In the future I will refactor this to move that page to the website. So we avoid these anti-patterns and don't create problems for the future.

zregvart:
@orpiske one more thing, the inline styles you have in `contents-table.html` the will not work on the production site, we explicitly disable inline CSS and JavaScript via [CSP](https://github.com/apache/camel-website/blob/034de87cf0f5ad866951f5d03af5e85d6c13d49a/static/.htaccess#L1384). To test the website in a production-like environment build with `CAMEL_ENV=production` and [run Apache HTTPD locally](https://github.com/apache/camel-website/blob/main/support/README.md#running-apache-http).

orpiske:
> @orpiske one more thing, the inline styles you have in `contents-table.html` the will not work on the production site, we explicitly disable inline CSS and JavaScript via [CSP](https://github.com/apache/camel-website/blob/034de87cf0f5ad866951f5d03af5e85d6c13d49a/static/.htaccess#L1384). To test the website in a production-like environment build with `CAMEL_ENV=production` and [run Apache HTTPD locally](https://github.com/apache/camel-website/blob/main/support/README.md#running-apache-http).

Thanks. It's looking like I might need to move those to the website repo and reference them there somehow. I'll investigate this more closely. Thanks for pointing this.

PR #947, opened 2022-12-17
Created release-3.19.0.md

valdar:
Camel Kafka Connector 3.19.0 release notes.


davsclaus (changes requested):
We do not support java 8, so that should be removed. Camel 3.19 supports Java 11 and 17. So I assume you should change it to that.

oscerd (approved):

valdar:
> 



> We do not support java 8, so that should be removed. Camel 3.19 supports Java 11 and 17. So I assume you should change it to that.

Thanks. Done that.

davsclaus (approved):

PR #951, opened 2022-12-22
A blog about 2022 in numbers. Numbers need to be updated next week to…

davsclaus:
… catch up


davsclaus:
I grabbed our last 2019 blog and updated. Anyone is welcome to add more details to this if we have more information for 2022 that would be good to include.

squakez:
> I grabbed our last 2019 blog and updated. Anyone is welcome to add more details to this if we have more information for 2022 that would be good to include.

I will add something about Camel K, I was thinking to do a blog on its own but it makes sense to have one unique overview!

oscerd (approved):

davsclaus:
Well I think the number blog post should be a short post with just some stats.

And then we can do another post with project summary of 2022, where we can either have per project or a single post for all the projects together.

squakez:
@davsclaus I've removed my commit. I'll use that to spin off a new blog just for Camel K during next week.

squakez (approved):

PR #972, opened 2023-02-04
Karavan 3.20.0 release

mgubaidullin:


oscerd (approved):

davsclaus (approved):

mgubaidullin:
@davsclaus @oscerd I don't see my post on the preview? Any way to check the post that to be published on a future date?

davsclaus:
Yeah in the past we could see upcoming blogs, but that seems no longer possible. For now I think you then need to adjust the date to today or in the past, and then view, and then update before merging.

PR #995, opened 2023-04-21
Updated rules.js

PasunuriSrinidhi:
1. Refactored the HtmlTitle rule to extract the title validation into a separate function validateTitle for better organization and reusability.
2. Used a for loop instead of forEach for better performance in the RelativeLinks rule.
3. Removed the start variable in the StructuredData rule and instead used the textContent property of the closing script tag to extract the JSON-LD content.

### DIFF
--- a/rules.js
+++ b/rules.js
@@ -6,25 +6,29 @@ class HtmlTitle extends Rule {
       description: "Title of the page must be defined"
     };
   }
+  validateTitle(title) {
+    if (!title.textContent || title.textContent.trim().length === 0) {
+      this.report(title, "No title specified");
+    } else if (title.textContent.startsWith('-')) {
+      this.report(title, "Title starts with the dash `-`, add the `title` front matter to the Hugo content");
+    } else if (title.textContent.startsWith('Untitled')) {
+      this.report(title, "Title starts with the `Untitled` make sure that the Asciidoc file starts with level 1 heading");
+    }
+  }
+
 
   setup() {
     this.on("dom:ready", event => {
       const title = event.document.querySelector('html head title');
       if (title === null) {
         this.report(event.document.querySelector('html'), "Missing <title> tag in the <head>");
-      }
-      else if (!title.textContent || title.textContent.trim().length === 0) {
-        this.report(title, "No title specified");
-      }
-      else if (title.textContent[0] === '-') {
-        this.report(title, "Title starts with the dash `-`, add the `title` front matter to the Hugo content");
-      }
-      else if (title.textContent.startsWith('Untitled')) {
-        this.report(title, "Title starts with the `Untitled` make sure that the Asciidoc file starts with level 1 heading");
+      } else {
+        this.validateTitle(title);
       }
     });
   }
 }
+
 
 class RelativeLinks extends Rule {
   documentation(context) {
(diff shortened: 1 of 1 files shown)


rules.js
```diff
@@ -36,15 +40,14 @@ class RelativeLinks extends Rule {
...
     this.on("dom:ready", event => {
       const anchors = event.document.getElementsByTagName('a');
-      if (anchors === null) {
-        return;
-      }
-      anchors.forEach(a => {
-        const href = a.getAttribute("href");
-        if (href && href.value.startsWith("https://camel.apache.org")) {
-          this.report(a, `For links within camel.apache.org use relative links, found: ${href.value}`);
+      if (anchors !== null) {
```
zregvart, line 43: I prefer the non-nested variant, i.e.

```javascript
if (anchors === null) {
  return;
}
```

This [blog post](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html) articulates the reasons why.

rules.js
```diff
@@ -6,26 +6,30 @@ class HtmlTitle extends Rule {
       description: "Title of the page must be defined"
     };
   }
+  validateTitle(title) {
```
zregvart, line 9: If we're going with this approach, I'd prefer putting all the logic in the `validateTitle` function.

```javascript
validateTitle(event) {
//...
}
```

And below we would have:

```javascript
this.on("dom:ready", validateTitle)
```

rules.js
```diff
@@ -36,15 +40,14 @@ class RelativeLinks extends Rule {
...
-        return;
-      }
-      anchors.forEach(a => {
-        const href = a.getAttribute("href");
-        if (href && href.value.startsWith("https://camel.apache.org")) {
-          this.report(a, `For links within camel.apache.org use relative links, found: ${href.value}`);
+      if (anchors !== null) {
+        for (let i = 0; i < anchors.length; i++) {
+          const href = anchors[i].getAttribute("href");
+          if (href && href.startsWith("https://camel.apache.org")) {
```
zregvart, line 46: Missing `href.value` here

```suggestion
          if (href && href.value.startsWith("https://camel.apache.org")) {
```

rules.js
```diff
@@ -56,21 +59,12 @@ class StructuredData extends Rule {
...
-    this.on('tag:close', async event => {
-      const tag = event.previous;
-      if (event.target.nodeName === 'script' && tag.hasAttribute('type') && tag.getAttribute('type').value === 'application/ld+json') {
-        const startIdx = data.indexOf('>', start) + 1;
-        const endIdx = event.target.location.offset - 1; // omit the opening tag angled bracket
-        const content = data.substring(startIdx, endIdx);
+  
+setup() {
+    this.on('tag:close', event => {
+      const tag = event.target;
```
zregvart, line 65: For `tag:close` the event doesn't have a `target` property and that's why we need to remember the offset location of when the tag was opened. In the `tag:open` we also don't have access to attributes. This is why the logic here how it is.

zregvart (changes requested):
This needs a bit of work. To test try running the build (`yarn build`) to produce the content in `public/` and then you can test against `public/index.html` by running `yarn run html-validate public/index.html` for faster turnaround times.

rules.js
PasunuriSrinidhi: Thanks for your suggestion .I will edit the code

rules.js
PasunuriSrinidhi: Thanks for your review. I will write the entire logic in the validateTitle function

rules.js
PasunuriSrinidhi: Thanks for your help. I will add the href.value and modify the code accordingly

rules.js
PasunuriSrinidhi: Thank you so much for finding the errors. I will look into it.

PasunuriSrinidhi:
Updated the whole code. Can you please review the code and give suggestions? 
Thanks

rules.js
zregvart: I'm not seeing this change

rules.js
zregvart: This was not changed

zregvart:
@PasunuriSrinidhi thanks, this still needs a bit of work, see my other comments

PasunuriSrinidhi:
@zregvart , Thanks for your response. I will into all the comments and I will modify according to it .

Thank you

davsclaus:
@PasunuriSrinidhi let us know when you are done with the changes and we with help from @zregvart can take a look again

davsclaus:
closing old PR that was not active

PR #1065, opened 2023-09-20
Attempt to remove old docs for EOL releases

davsclaus:


oscerd (approved):

davsclaus:
Okay so this looks good, we now have updated docs for the LTS releases and removed old docs for EOL releases.

PR #1097, opened 2023-11-24
feat: add SBOM downloads based on cut-off date

### LINKED ISSUE #1096: Showing SBOM files in the download page
Starting from 4.1.0 we are providing SBOM for some of our projects:

- Camel
- Camel-Quarkus
- Camel-Spring-Boot
- Camel-Kamelets

I think it would be good to show these files in the Download page.

The files are json and XML

https://repo1.maven.org/maven2/org/apache/camel/camel/4.1.0/ 

For example.

@zregvart do you think it is something we could add? We'll need to filter out some of the versions for this subprojects and only show the columns in particular versions.

zregvart:
With this the SBOM downloads will be available for any release created after 2023-10-11. Both JSON and XML download links will be available. The Apache conventions for distributing these via dist.apache.org must be followed, so .asc and .sha512 files must be present in addition to the .json and .xml SBOM files.

Resolves #1096


zregvart:
@oscerd PTAL :)

oscerd (approved):
Looks good to me. 

I guess there is some difference around the naming conventions:

For example for camel-quarkus is

camel-quarkus-3.5.0-cyclonedx.json

While for Spring-boot is

spring-boot-4.2.0-cyclonedx.json

This is not a real problem because we can renominate the files will creating the dist/dev folder and signing them.

oscerd:
This looks really great, I will align the naming convention around and coordinate the dist/dev push and dist/release post vote process.

oscerd:
Thanks a lot @zregvart !

zregvart:
The, semi-obvious drawback of using the cutoff date is that all releases (in the select categories) made from that date forward need SBOMs. That includes any 3.21.x, 4.0.x and 4.1.x patch releases.
We also do not offer separate downloads for Camel Spring Boot, so I'm not sure how to handle that with this approach.

zregvart:
Let's also wait for the preview and see what cutoff date we would like to set before merging this...

davsclaus:
We can wait until January 2024, then 3.14 and 3.20 is EOL and no new releases of those.
Then its only 3.21 and 2.22 for 3.x. And maybe you can skip those, so it will be only v4 onwards

oscerd:
Let's try to prepare everything so eventually it will be easy to publish these informations.

oscerd:
Another possibility is adding a separated page for SBOMs, or maybe create a table in Security page.

oscerd:
For 4.3.0, coming release, we should be fine, I added the steps in release-guide, same for camel-quarkus and camel-kamelets. @zregvart can we maybe add some kind of "N/A" for releases 3.21.x (and the 3.x in general) and 4.0.x and 4.1.x? So we can try to publish what we have.

oscerd:
Pushed for 4.0.3 too

oscerd:
Looks good. I'll need to add the sbom profile to release also for 3.21.x and ckc I guess, but for the moment it looks good.

PR #1098, opened 2023-11-24
Added new blog post about the performance improvements from 4.0 to 4.2

orpiske:


davsclaus (approved):

davsclaus:
I guess the bug in 4.1.0 and 4.0.2 (I assume) shows the degration in your graph

orpiske:
> I guess the bug in 4.1.0 and 4.0.2 (I assume) shows the degration in your graph

That's correct. The impact of not caching the type converter was pretty significant in that scenario. I was a great finding by @aldettinger who noticed it when running their performance tests.

orpiske:
Obs.: the images are inverted on the post. I had fixed it yesterday, but a linking error prevented the blog post from being regenerated. The new one should contain this fix and a few other things I found.

aldettinger:
@orpiske Really nice kind of article. It's really awesome to be transparent about the test setup and methodologies :+1:

PR #1121, opened 2024-01-10
chore: update GitHub actions

zregvart:
Main motivation here is to get faster artifact uploads, which currently take up significant time of the workflow.


zregvart:
The preview workflow is running with the versions from `main`, i.e. not with the upgraded dependencies. I'll merge this, if there are any issues I'll do a followup PR. I think other PRs will need to be rebased for the preview to function.

PR #1135, opened 2024-01-30
Several dependency updates

zregvart:


zregvart:
The netlify-cli is broken for us right now (updated in an earlier PR), I think I'll revert to an older version (in a separate PR). Other than that changes here look good. So I'm merging this now.

PR #1188, opened 2024-06-03
Create release-4.4.2.md

valdar:


davsclaus:
I wonder if Kafka Connector can run on JDK21 ?
As Camel 4.4 can do that, and if so it should be 17,21

valdar:
It seems so : https://github.com/apache/camel-kafka-connector/actions/runs/9370239202

valdar:
I have added the lts kind as well.

davsclaus (approved):

PR #1307, opened 2025-02-18
feat: Camel K release 2.6.0 announcement

squakez:


oscerd (approved):

davsclaus:
Can you remove the doc changes (2.4/2.6) as we have problems with building the website. You can get the blog ann publiushed first, then afterwards add the 2.6 docs. And then we need to find a way to remove the old docs and make this better in the future where camel-k is more independent from kamelets / quarkus et all and not create this big mess.

squakez:
> Can you remove the doc changes (2.4/2.6) as we have problems with building the website. You can get the blog ann publiushed first, then afterwards add the 2.6 docs. And then we need to find a way to remove the old docs and make this better in the future where camel-k is more independent from kamelets / quarkus et all and not create this big mess.

Sure, I'll have a look. However I think it is more the other way around. There is a dependency from Kamelets toward Camel K which we need to address to make all modules independent.

davsclaus (approved):

Issue #1176, opened 2024-04-23
Overlapping UI on the Blog Page


aashnajena:
The topics list on the left side seems to overlap with the footer. The height of the container having that list needs to be fixed.

<img width="1506" alt="Screenshot 2024-04-23 at 5 34 33 PM" src="https://github.com/apache/camel-website/assets/32356795/1cf82a49-4790-4528-8a1d-e812ae18da80">

Suvendu-UI:
can i contribute to this issue ?

zregvart:
@Suvendu-UI you're very welcome to propose a change via a pull request, have a look at the [contribution guide](https://github.com/apache/camel-website?tab=readme-ov-file#contribute-changes).

ONKARBH:
@zregvart ,
hello sir,

I want to work on this issue,
If you don't mind i will work on this issue

tadayosi:
@ONKARBH You can go ahead to work on a PR and submit it.

Note as the policy we don't prefer annonymous GitHub profiles because of security and trustness:
https://camel.apache.org/camel-core/contributing/expectations.html
If you consider contributions please follow our guideline as well.

Abhi-2516:
@zregvart Hello Sir, if the issue still open to work , I want to work on it .

ONKARBH:
Hello @tadayosi, @zregvart,

I have fixed the overlapping UI issue on the Blog Page sidebar as mentioned in issue #1176.
The fix ensures that the topics list does not overlap with the footer by adjusting the container height and making it scrollable when necessary.

I have submitted a Pull Request #1313 for review. Kindly check it at your convenience.
Please let me know if any modifications are needed. Looking forward to your feedback.
Thank you.

![Image](https://github.com/user-attachments/assets/c99838f4-a2f8-4d6d-b504-a9200ae54cb8)

PR #1313, opened 2025-02-20
 Fix Overlapping UI on Blog Page Sidebar

ONKARBH:
I was resolved issue #1176, where the topics list on the left side of the Blog Page was overlapping with the footer.

Screenshot 2025-02-20 115914


davsclaus:
there is something wrong with this PR with 5000+ changed files. Can you try again and only send the actual changes needed and not some cache updates.

PR #1340, opened 2025-04-17
Update antora-playbook.yml added ckc 4.11.x

valdar:


oscerd (changes requested):
We should show only LTS. 4.11.x is dev.

PR #1355, opened 2025-06-23
feat(blog): Camel K 2.7.0

squakez:


davsclaus:
The `NOTE:` is not shown nicely as that is for adoc - for md you need to do something differently

PR #1358, opened 2025-06-24
chore: switch from Yarn PnP to node-modules

zregvart:
This changes the Yarn linker method from PnP, where all dependencies are placed in the git repository to the regular node-modules linker, much akin to how most Node projects handle dependencies.

The PnP linker mode has served us very well, with only a few issues along the way, namely the fact that the architecture-specific binaries would be cached. This meant that the state of the git repository would change depending on where the build, or update of the dependencies was done.

Using the node-modules doesn't exhibit those issues as the dependencies are fetched either from the network or from the local per-user cache.

The other side effect is that the checked out size is significantly smaller compared to PnP.


zregvart:
I'm going to merge this, I'm optimistic that this will not cause any issues, if it does I'll revert as soon as I'm aware any issues.

PR #1359, opened 2025-06-24
feat: quick lookup for kamlet catalog

### LINKED ISSUE #1354: Kamelets Catalog - Make the catalog searchable
From https://github.com/apache/camel-kamelets/issues/2503

The Kamelets Catalog reached 250 Kamelets. Congrats for that!

The downside for so many Kamelets is that it becomes it bit hard to find Kamelets.

The Camel component reference:

https://camel.apache.org/components/4.10.x/

Has a Quick lookup. This would also be convenient for the Kamelets catalog.

zregvart:
Adds the quick lookup for kamlet catalog, reusing the same logic that the component reference already uses.

Fixes #1354


davsclaus (approved):

zregvart:
Looks okay in the preview.

PR #1373, opened 2025-08-14
Improve documentation navigation

christophd:
- Improve reference doc navigation with css styles
- Proper sublist indent
- Always highlight top level categories (components, dataformats, languages, eips)
- Display top level categories in filtered search results

## Top Level Categories

One of the biggest issues in reference doc navigation was that different top level categories have not been highlighted in the navigation. As seen in this example where `Data Formats` is just like an arbitrary nav item.

reference_nav_sublist

The update highlights the categories in the navigation so users can see that a new category begins in the navigation:

reference_nav_sublist_update

## Sublists

Also, sublists of Camel components are displayed on the same navigation level as seen in this example about `AI` components (chatscript, kserve, ...):

reference_nav

This has been improved with proper indentation of sublists:

reference_nav_update

## Search results

When searching the navigation mixed the different categories, sometimes producing duplicates when component and dataformat results match.

search_filter

Now top level categories are also displayed in the search results for better separation

search_filter_update

## User Manual

The user manual navigation also benefits from the updates:

user_manual

user_manual_update

## User Manual (expanded)

user_manual_expanded

user_manual_expanded_update

## Sublist naming

As a follow-up task I suggest to remove the duplication in sublist component names, like seen in AWS* components (skip the AWS prefix in the sublist).

aws_sublist_naming

### DIFF
--- a/antora-ui-camel/src/css/nav.css
+++ b/antora-ui-camel/src/css/nav.css
@@ -128,16 +128,46 @@ html.is-clipped--nav {
   padding: 0;
 }
 
-.nav-panel-menu input.search:valid ~ .nav-menu .filtered {
-  display: none;
-}
-
 .nav-panel-menu input.search:valid ~ .nav-menu button {
   display: none;
 }
 
-.nav-panel-menu input.search:valid ~ .nav-menu a.nav-link:not(.filtered) {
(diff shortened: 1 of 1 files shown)


orpiske:
Wow, that's amazing!

Issue #1137, opened 2024-02-01
Copy button for source code blocks


jamesnetherton:
Quite a few projects have a button / icon in source code snippets to copy the content onto the clipboard. It'd be nice if we can do the same for the Camel docs. It makes life simpler when experimenting in an IDE or on the command line.

PR #1398, opened 2025-10-03
feat: add copy to clipboard style

gansheer:
Ref #1137 

The copy button already existed on the antora default ui (https://docs.antora.org/antora-ui-default/code-blocks/#copy-to-clipboard). I added the functionality again (with some adaptations).

One thing to point out is about console blocks: https://docs.antora.org/antora-ui-default/code-blocks/#console-blocks.
I will probably need to change the existing documentation but it manages the sh commands pretty well

It looks like this on the demo preview:

https://github.com/user-attachments/assets/292f1655-8423-48cc-a139-e3be8dc85a2f

### DIFF
--- a/antora-ui-camel/src/css/doc.css
+++ b/antora-ui-camel/src/css/doc.css
@@ -616,8 +616,108 @@
   right: 0.25rem;
 }
 
+/* Hide the old language display since we now use source-toolbox */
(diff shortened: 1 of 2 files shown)


davsclaus:
Thanks this is nice

Issue #1281, opened 2025-01-25
Make search box only index and search core camel docs


davsclaus:
When using the search box, then it should only search the camel-core project as users look for general Camel information.
Today it shows CK, CEQ first over core camel.

The sub projects have special docs (limited) that are special for CSB CEQ that users can browse and see.

shuhaki:
can you assign me..... i want to do

oscerd:
I assigned this to you. If you don't work on this, just let us know

davsclaus:
@christophd see this ticket

gansheer:
@davsclaus Does that means we would not be able to search in the subprojects documentation ?

davsclaus:
> [@davsclaus](https://github.com/davsclaus) Does that means we would not be able to search in the subprojects documentation ?

yes, for sub projects they have little docs, and you just browse their docs. 99% of times you need to read the core docs.

Sumit20-4:
take

PR #1456, opened 2026-01-02
Fix: Restrict search to camel-core docs index and prioritize core results

GaneshPatil7517:
This PR updates the documentation search behavior so that queries return results **only from the `camel-core` documentation index**, filtering out sub-project docs such as CK, CEQ, or CSB CEQ that previously appeared with higher priority.  
The goal is to improve relevance for users who expect the main search bar to surface **general Camel core information first**.

### What was changed
- Search scope is now limited to the `camel-core` docs index.
- Sub-project documentation is ignored in global search results.
- Core Camel documentation results are ranked and displayed first.
- Existing search features and UI behavior remain intact.

###  Testing
- Verified search results return only camel-core matches.
- Confirmed result ordering prioritizes core docs.
- No console errors, regressions, or performance issues observed.

###  Notes
This change improves user experience by ensuring global search stays focused on the core documentation set, while still allowing sub-project docs to be browsed manually in their respective sections.

Closes: #1281

### DIFF
--- a/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
+++ b/antora-ui-camel/src/js/vendor/algoliasearch.bundle.js
@@ -10,7 +10,7 @@
     '/camel-spring-boot/',
     '/camel-kafka-connector/',
     '/camel-kamelets/',
-    '/camel-karaf/'
+    '/camel-karaf/',
   ]
 
   // Check if a URL belongs to a sub-project that should be filtered out


davsclaus:
This PR also includes the previous PR for maven wrapper. Can you make a clean PR with only search changes

GaneshPatil7517:
hey @davsclaus See now link:- https://github.com/apache/camel-website/pull/1457

PR #1476, opened 2026-01-16
Redirect rule, when the header Accept: text/markdown is set, than red…

Croway:
…irect to the .md page

Add Accept: `text/markdown` content negotiation for LLM access.

Requests with `Accept: text/markdown` header are redirected (303) to the corresponding .md file.

Example:
```
curl -H "Accept: text/markdown" .../direct-component.html → 303 redirect to direct-component.md
```

For reference:
```
  [NC] - No Case
  - Makes the pattern match case-insensitive
  - So text/markdown, Text/Markdown, TEXT/MARKDOWN all match

  [R=303,L] - Two flags combined:
  - R=303 - Redirect with HTTP status code 303 (See Other)
  - L - Last rule
    - Stop processing further rewrite rules if this one matches
    - Without this, Apache continues checking subsequent rules
```

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -1,6 +1,12 @@
 # Redirect security advisories to new location
 RewriteEngine on
 RewriteBase "/"
+
+# Content negotiation for LLM-friendly Markdown
+# When Accept header requests text/markdown, redirect .html to .md
+RewriteCond %{HTTP_ACCEPT} text/markdown [NC]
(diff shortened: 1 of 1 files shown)


squakez (commented):
It may not work with fragments, but I guess it is not a concern for LLMs. Or is it?

davsclaus (approved):

PR #1479, opened 2026-01-16
Add WARNING section when documentation is marked as deprecated

### LINKED ISSUE #1477: Add WARN section if a component is deprecated
In the top of the doc page there are some metadata

https://github.com/apache/camel/blob/main/components/camel-olingo2/camel-olingo2-component/src/main/docs/olingo2-component.adoc?plain=1#L8

If there is a `:deprecated` then this page should automatic have a 

WARNING: This :shortname is deprecated

Where :shortname is also in the metadata, as this can also apply for languages, data formats and misc

GaneshPatil7517:
This PR implements automatic rendering of a WARNING section on documentation pages when the page metadata marks the component (or language, data format, misc) as deprecated.

When a documentation page includes the `:deprecated:` attribute in its AsciiDoc header, a prominent warning block will automatically appear at the top of the page (below the title), alerting users that the documented component/feature is deprecated.

## Changes

Modified `antora-ui-camel/src/partials/article.hbs` to conditionally render a deprecation warning block:

- Checks if the page has the `deprecated` attribute set
- Renders an AsciiDoc-style WARNING admonition block with consistent styling
- Uses the component's `:shortname:` if available, otherwise falls back to the page title
- No manual changes required in individual AsciiDoc files
- Applies to all documentation types: components, languages, data formats, misc

## Implementation Details

The solution:
1. Uses Handlebars' `{{#unless}}` block to check if the deprecated attribute is undefined
2. Renders the warning using the existing admonition block styling (`.admonitionblock.warning`)
3. Follows the Antora/AsciiDoc standard table structure for admonitions
4. Leverages the `{{or}}` helper to display shortname or fallback to page title

## Example

For a page with:
```asciidoc
= My Component
:shortname: mycomponent
:deprecated: true
```

The rendered page will display:
```
┌──────────────────────────────────────────┐
│ ⚠️  Warning                               │
│ This mycomponent is deprecated            │
└──────────────────────────────────────────┘
```

## Testing

Locally tested:
- Built the Antora UI and website with the template changes
- Verified that pages with `:deprecated: true` show the warning
- Confirmed non-deprecated pages remain unchanged
- CSS styling from existing admonition block styles is applied correctly

## References

Closes #1477

### DIFF
--- a/antora-playbook-snippets/antora-playbook.yml
+++ b/antora-playbook-snippets/antora-playbook.yml
@@ -106,7 +106,7 @@ asciidoc:
 runtime:
   log:
     level: warn
-    failure_level: error
+    failure_level: warn
 #    format: json
 
 antora:


oscerd (approved):

davsclaus:
I do not see that deprecated warning such as
https://pr-1479--camel.netlify.app/components/4.14.x/nitrite-component.html

davsclaus (changes requested):
this is wrong as it edits an article.hbs file which is not related to camel components. I think this should be done in camel-core/docs folder, where the components / languages et all are

davsclaus:
Screenshot 2026-01-19 at 08 13 27

The text that says is deprecated should be shown with some kind of highlight

For example like the WARNING you can see if you scroll down this page
https://camel.apache.org/components/4.14.x/zookeeper-component.html

PR #1570, opened 2026-04-13
Remove Camel 3.x docs

davsclaus:


antora-ui-camel/public/_/partials/footer-content.hbs
```diff
@@ -24,7 +24,7 @@
                 <dd><a href="{{siteRootPath}}/camel-kafka-connector/next/">Camel Kafka Connector</a></dd>
                 <dd><a href="{{siteRootPath}}/camel-quarkus/next/">Camel Quarkus</a></dd>
                 <dd><a href="{{siteRootPath}}/camel-spring-boot/next/">Camel Spring Boot</a></dd>
-                <dd><a href="{{siteRootPath}}/camel-karaf/3.22.x/">Camel Karaf</a></dd>
+                <dd><a href="{{siteRootPath}}/camel-karaf/latest/">Camel Karaf</a></dd>
```
gansheer, line 27: Out of curiosity, is there a reason why karaf uses latest instead of next ?

davsclaus: I dont know - i think it may be that they didnt migrate to next style or something - when using next it fails, and latest works.

Issue #1410, opened 2025-10-09
Changes not deployed by jenkins build


gansheer:
There is a strange behavior with the jenkins build: when deploying after the new build it does not pick up the changed files.
In the https://github.com/apache/camel-website-pub branch `asf-site` we can see that the generated commits keep overriding the changes made to the file .asf.yml like this PR https://github.com/apache/camel-website-pub/pull/2.

The step in charge of this is:
https://github.com/apache/camel-website/blob/d2ba92b1fe6e1904540b48abacca4c26a1cd344b/Jenkinsfile#L93-L102

Every time it is applyed we end up with the old value from the current repo: https://github.com/apache/camel-website/blob/d2ba92b1fe6e1904540b48abacca4c26a1cd344b/.asf.yaml#L28

PR #1573, opened 2026-04-15
INFRA-27679 - Avoid squash to avoid new very large commit on each new

apupier:
website build

the newly created very large commit created on each push might be the cause of the 404 as the CDN might have problems to handle it right away. Based on https://issues.apache.org/jira/browse/INFRA-27679?focusedCommentId=18062765&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18062765

It means that the history will grow. So we might need to ensure that the checkout of this site is either done with a depth of 1 or if already available and just a pull is done maybe to squash one time manually from time to time.
see https://issues.apache.org/jira/browse/INFRA-27679?focusedCommentId=18073130&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18073130

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,7 +17,7 @@
  * under the License.
  */
 def NODE = 'git-websites'
-def STOP_SQUASH_AT = '3381ad5637eb525502df319e84b7208e8f2a977b'
+def VALID_ASF_YAML = '3381ad5637eb525502df319e84b7208e8f2a977b'
 
 pipeline {
     agent {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -93,16 +92,10 @@ pipeline {
                 dir('deploy/live') {
                     deleteDir()
                     sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website-pub.git .'
-                    sh "git -c core.editor='sed -i 2,/\$(git log --skip=9 -1 --pretty=format:%h)/s/^pick/squash/' rebase -q --interactive $STOP_SQUASH_AT" // squash all but initial and last 9 commits
                     sh 'git rm -q -r *'
                     sh "cp -R $WORKSPACE/camel-website/public/. ."
                     sh 'git add .'
-                    sh "git checkout $STOP_SQUASH_AT -- ./.asf.yaml" // force revert to commit containing the valid .asf.yml
```
apupier, line 100: need to check the content of the .asf.yaml which will be provided then, maybe will need to change something how the .asf.yaml is genrated or to continue to checkout one from a specific revision.
I'm wondering if the problem is not that we wan a different .asf.yaml between camel-website repo and camel-website-pub

apupier: @gansheer as you introduced the line for that, maybe you can provide a bit more context?

davsclaus (approved):

davsclaus:
@zregvart ^^^ in case you want to take a quick look

zregvart (approved):
Let's give it a shot, I don't think there is much that this can break if anything. Also easy to revert if it ends up not helping.

Thanks @apupier

Jenkinsfile
gansheer: I only modified that line but yes there is context: https://github.com/apache/camel-website/issues/1410#issuecomment-3390139034. From what I remember we squash every commit to avoid the repo to grow too much (I would say in storage and history). But since we want to keep the .asf.yaml file from the camel-website-pub repo (and not the one from the camel-website repo) we retrieve the valid version from a defined commit we decided contained the right version. 

I think with your change we will end with camel-website .asf.yml file inside camel-website-pub so it probably won't work well.

camel-website: https://github.com/apache/camel-website/blob/main/.asf.yaml
camel-website-pub: https://github.com/apache/camel-website-pub/blob/asf-site/.asf.yaml

apupier: ok, so I changed the PR to keep the .asf.yaml from the camel-website-pub

PR #1633, opened 2026-06-11
Reduce deploy diff size with rsync and add CDN nudge push

### LINKED ISSUE #1631: Reduce website size and fix CDN 404s — root cause analysis and action plan
## Summary

Deep analysis of the sporadic 404 errors reported in [INFRA-27679](https://issues.apache.org/jira/browse/INFRA-27679) and #1565. Pages that work fine suddenly return 404 hours later — in browsers only, not curl. The root cause is Fastly caching 404 responses that it should never cache.

### Why working pages suddenly 404

Here is what happens, step by step:

1. **Page is cached and working** — Fastly CDN has a valid cached copy, users see the page fine
2. **CDN cache entry expires** — Fastly's TTL for this page runs out (hours later)
3. **CDN re-fetches from origin** — Fastly sends a request to the Apache origin server to refresh the cache
4. **Origin returns 404 during deploy sync** — the Jenkinsfile deploy does \`git rm -r * && cp -R public/ .\` then pushes. ASF infra syncs this to the web server. During that sync window, files may briefly not exist on the origin
5. **CDN caches the 404** — Fastly treats the 404 as a valid response and caches it with its own TTL
6. **Page is broken for all browser users** — browsers send \`Accept-Encoding: gzip\`, and the gzip cache variant now holds a cached 404. Curl works because it doesn't send gzip by default, hitting a different cache variant (\`Vary: Accept-Encoding\` splits the cache)

**This is a CDN configuration problem, not a website content problem.** No CDN should cache 404 responses from an origin. The page exists, the origin just returned 404 transiently during a deploy window.

#### 1. Fastly CDN caches 404 responses (the trigger)

@gansheer's testing (#1565) confirmed the 404 **only** occurs with \`Accept-Encoding: gzip\`. Browsers always send gzip — curl does not by default. \`Vary: Accept-Encoding\` causes Fastly to maintain separate cache entries per encoding. The gzip variant gets a cached 404 while the non-gzip variant may be fine. @zregvart confirmed the origin returns 200 — the 404 is purely in the CDN cache.

#### 2. Deploy creates a window where origin returns 404 (the cause)

~~The Jenkinsfile deploy did \`git rm -r * && cp\` which deleted and re-added every file, creating a window where files didn't exist on origin during sync.~~ **Fixed in PR #1633** — deploy now uses \`rsync --delete\` so only actually-changed files are modified, eliminating the all-files-deleted window.

#### 3. CDN purge fails silently for large pushes (the amplifier)

~~Because \`git rm -r * && cp\` marked every file as changed, the push contained ~15,000+ file changes. The PURGE event payload exceeded the 256KB message queue limit and was silently dropped.~~ **Fixed in PR #1633** — rsync means git only sees actually-changed files, plus a follow-up "nudge" push ensures the PURGE fires.

### Full rebuild still triggers 14K+ file changes

Even with rsync, a **full rebuild** (all ~14K files changed) can still happen when a global Hugo template is modified. For example, PR #1639 changed `header.html` (included on every page), which caused Hugo to regenerate all ~14K HTML files with different `og:image` meta tags. The `build:markdown` step then also regenerated all `.md` files from the changed HTML. This pushes a large git diff through rsync and may again overflow the 256KB PURGE queue — causing some 404s until the CDN cache expires naturally.

**This is unavoidable for template changes** — the only real fix is the INFRA VCL change to stop caching 404s.

## The fix: INFRA must stop Fastly from caching 404s

**This is the single most important remaining action item.** In Fastly VCL:
\`\`\`
sub vcl_fetch {
  if (beresp.status == 404) {
    set beresp.cacheable = false;
  }
}
\`\`\`

@zregvart already suggested this on the INFRA Jira ticket. This one change would eliminate the entire class of problems — even if the origin briefly returns 404 during a deploy, the CDN would pass it through without caching, and the next request would get the real page.

## Pull Requests

| PR | Description | Savings | Status |
|---|---|---|---|
| #1639 | Blog images: shared default, dedup, resize, PNG→JPEG | **~422 MB** / 414 files deleted, 229 optimized | **Merged** |
| #1640 | Kafka-connector docs: main branch only | ~48 MB / ~900 files | **Merged** |
| #1633 | Jenkinsfile: rsync deploy + CDN nudge push | only changed files in deploy diff | **Merged** |
| #1634 | Remove mod_deflate and ETag stripping | removes unnecessary Apache directives | **Merged** |
| #1645 | Parallelize build post-steps and checks | faster CI wall-clock time | Open |

**Merged so far: ~470 MB site reduction + rsync deploy + .htaccess cleanup.**

~~PR #1635 (Remove EOL 1.x/2.x schema files)~~ — **Closed.** The schema XSD files are not included in any source JAR, so the website is the only place they are hosted. They need to stay.

### Critical: INFRA VCL fix
Stop Fastly from caching 404 responses. This is tracked in [INFRA-27679](https://issues.apache.org/jira/browse/INFRA-27679).

### Medium: Further Antora reduction
| Sub-project | 3-version size | Notes |
|---|---|---|
| camel-components | 317 MB | core docs — keep |
| camel-kamelets | 65 MB | could reduce to 2 versions |
| camel-quarkus | 69 MB | could reduce to 2 versions |
| camel-k | 28 MB | could reduce to 2 versions |

### Long-term
- **Remove 3.x documentation** — tracked in #1302 / PR #1570
- **Split static assets from generated content** — host from separate origin
- **Incremental Antora builds** — only regenerate changed components

## References

- [INFRA-27679](https://issues.apache.org/jira/browse/INFRA-27679) — Jira issue with INFRA team discussion
- #1565 — GitHub issue with gzip/Accept-Encoding findings
- #1302 — Remove 3.x docs
- #1533 — Empty commit CDN flush
- #1573 — Avoid squash to avoid large commits

davsclaus:
## Summary

- Use `rsync` instead of `git rm -r * && cp -R` in the deploy stage so git only sees actually-changed files rather than marking all ~15,000+ files as changed on every build
- Add a follow-up empty "nudge" commit + push after the main deploy to reliably trigger CDN cache purge

## Why

Per [INFRA-27679](https://issues.apache.org/jira/browse/INFRA-27679), the ASF message queue has a **256KB event payload limit**. Large pushes with 15,000+ changed files exceed this limit, causing the CDN PURGE request to be **silently dropped**. The CDN then serves stale content (including cached 404s) indefinitely.

The `rsync --delete` approach only produces git diffs for files that actually changed. A blog-only update might change 10 files instead of 15,000, staying well within the 256KB payload limit.

The follow-up nudge push (after a 5-second delay) provides a safety net: even if the main push exceeds the limit, this tiny second push will trigger the CDN purge.

## Test plan

- [ ] Verify rsync is available in the Jenkins Docker build environment
- [ ] Test that `.asf.yaml` is still correctly restored from `$VALID_ASF_YAML`
- [ ] Confirm the nudge push triggers CDN cache invalidation
- [ ] Monitor next deploy for 404 recurrence

Fixes #1631

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,11 +93,13 @@ pipeline {
                 dir('deploy/live') {
                     deleteDir()
                     sh 'git clone -b asf-site https://gitbox.apache.org/repos/asf/camel-website-pub.git .'
-                    sh 'git rm -q -r *'
-                    sh "cp -R $WORKSPACE/camel-website/public/. ."
-                    sh 'git add .'
+                    sh "rsync -a --delete --exclude='.git' --exclude='.asf.yaml' $WORKSPACE/camel-website/public/ ."
+                    sh 'git add -A'
                     sh "git checkout $VALID_ASF_YAML -- ./.asf.yaml" // force revert to commit containing the valid .asf.yml
-                    sh 'git commit -m "Website updated to $GIT_COMMIT"'
+                    sh 'git diff-index --quiet HEAD || git commit -m "Website updated to $GIT_COMMIT"'
+                    sh 'git push origin asf-site'
+                    sh 'sleep 5'
+                    sh 'git commit --allow-empty -m "CDN cache invalidation"'
                     sh 'git push origin asf-site'
                 }
             }


oscerd (approved):

davsclaus:
lets give this a go, as it may trigger a CDN cache purge by us

PR #1642, opened 2026-06-11
Adjust categories so it's not hidden along with Camel releases

orpiske:


gansheer:
@davsclaus  @orpiske If in the end we don't want to hide the release blog post maybe we should rollback the change ?

orpiske:
> @davsclaus @orpiske If in the end we don't want to hide the release blog post maybe we should rollback the change ?

This is more like a community / use case post ... not so much of a Camel release announcement per se, so IMHO, it should be fine - but I can rollback if others disagree.

davsclaus:
> > @davsclaus @orpiske If in the end we don't want to hide the release blog post maybe we should rollback the change ?
> 
> This is more like a community / use case post ... not so much of a Camel release announcement per se, so IMHO, it should be fine - but I can rollback if others disagree.

No this is what we should do much more - use-case how-tos etc, and keep release announcments smaller and just to make a heads up its released, and then cover topics in individual blog posts

PR #1700, opened 2026-07-21
Delegate release manifest generation to vendored WebsiteManifestGenerator.java

ammachado:
## Summary
- Replace `update-camel-cli-manifest.sh`'s hand-rolled checksum computation and immutability/forward-only manifest rules with a call into a vendored, byte-identical copy of apache/camel's own `WebsiteManifestGenerator.java` (`scripts/WebsiteManifestGenerator.java`, sourced from `dsl/camel-jbang/camel-launcher/src/jreleaser/java/WebsiteManifestGenerator.java`).
- Keep `static/install.sh` and `static/install.ps1` synced with apache/camel main (surfaced wget/tar errors, self-update pin-state tracking), and have the generator write `static/install.sha256` so `camel self-update` can verify the installer before running it.
- Carve `install.sh`/`install.ps1` out of the site's sensitive-file-extension `.htaccess` block rule so the documented `curl ... install.sh | sh` one-liner isn't 403'd.
- `update-camel-cli-manifest.sh` step 1 now also fetches `WebsiteManifestGenerator.java` from apache/camel main and reapplies the "VENDORED COPY" banner, so the generator itself stays mechanically in sync instead of requiring a manual copy-paste.
- This keeps the manifest format/rules and installer scripts in one place instead of reimplementing them in shell, so the website and the upstream tool can't drift apart.

## Test plan
- [ ] Run `./update-camel-cli-manifest.sh <version> --latest` against a real release and confirm `static/camel-cli/releases/<version>.properties`, `latest.properties`, and `static/install.sha256` are written correctly
- [ ] Verify immutability: re-running for an existing version with unchanged checksums is a no-op; a checksum conflict fails loudly
- [ ] Verify forward-only: attempting to move `latest.properties` backward fails loudly
- [ ] Confirm `curl https://.../install.sh | sh` is not blocked by `.htaccess`
- [ ] Confirm the script's new generator-sync step writes a `scripts/WebsiteManifestGenerator.java` byte-identical (banner aside) to apache/camel's source

### DIFF
--- a/scripts/WebsiteManifestGenerator.java
+++ b/scripts/WebsiteManifestGenerator.java
@@ -1,0 +1,320 @@
+// VENDORED COPY - do not edit here.
+// Source of truth: apache/camel
+//   dsl/camel-jbang/camel-launcher/src/jreleaser/java/WebsiteManifestGenerator.java
+// To update: re-copy that file over this one, then re-apply this banner. The body below is
+// kept byte-identical to the upstream source so the sync stays a mechanical copy.
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Intentionally in the default package so camel-package.sh can run this JDK-only source-file
+// tool directly by path without compiling or loading a package-qualified launcher class.
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.HexFormat;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * JDK-only tool that computes SHA-256 checksums for a release TAR/ZIP pair and writes the
+ * website's four-field release manifest ({@code format}, {@code version}, {@code tar_sha256},
+ * {@code zip_sha256}). Per-version manifests are immutable; {@code latest.properties} can only
+ * move forward to a higher semantic version and cannot silently change checksums for the same
+ * version. Writes are atomic.
+ *
+ * <p>
+ * Usage: {@code java WebsiteManifestGenerator.java --version X.Y.Z --tar <path> --zip <path>
+ * --output <dir> --latest <true|false>}
+ */
+public class WebsiteManifestGenerator {
+
+    private static final Pattern VERSION_PATTERN = Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)$");
+    private static final Set<String> REQUIRED_OPTIONS
+            = new LinkedHashSet<>(List.of("--version", "--tar", "--zip", "--output", "--latest"));
+    private static final String MANIFEST_FORMAT = "1";
+
+    // Properties-style ASF license header prepended to every generated manifest. These '#' comment
+    // lines carry no data and are skipped by this tool's own strict re-parser (parseStrictManifest)
+    // and by the installers (install.sh / install.ps1), which tolerate a commented manifest.
+    private static final String LICENSE_HEADER
+            = "## ---------------------------------------------------------------------------\n"
+              + "## Licensed to the Apache Software Foundation (ASF) under one or more\n"
+              + "## contributor license agreements.  See the NOTICE file distributed with\n"
+              + "## this work for additional information regarding copyright ownership.\n"
+              + "## The ASF licenses this file to You under the Apache License, Version 2.0\n"
+              + "## (the \"License\"); you may not use this file except in compliance with\n"
+              + "## the License.  You may obtain a copy of the License at\n"
+              + "##\n"
+              + "##      http://www.apache.org/licenses/LICENSE-2.0\n"
+              + "##\n"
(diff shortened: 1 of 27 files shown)


scripts/WebsiteManifestGenerator.java
```diff
@@ -0,0 +1,320 @@
+// VENDORED COPY - do not edit here.
+// Source of truth: apache/camel
+//   dsl/camel-jbang/camel-launcher/src/jreleaser/java/WebsiteManifestGenerator.java
+// To update: re-copy that file over this one, then re-apply this banner. The body below is
+// kept byte-identical to the upstream source so the sync stays a mechanical copy.
```
davsclaus, line 5: This banner says "kept byte-identical to the upstream source," but the file below has three functional changes vs the current upstream:

1. Added `LICENSE_HEADER` constant (not in upstream)
2. `renderManifest()` prepends `LICENSE_HEADER` (upstream doesn't)
3. `parseStrictManifest()` adds `line.startsWith("#")` skip (upstream only skips empty lines)

A future sync that follows this banner literally ("re-copy, re-apply banner") would silently lose these changes.

update-camel-cli-manifest.sh
```diff
@@ -74,58 +56,25 @@ for script in install.sh install.ps1; do
     echo "  wrote static/$script"
```
davsclaus, line 61: This loop downloads `install.sh` and `install.ps1` from upstream `apache/camel main` and overwrites the local copies. Since the upstream versions do **not** have the comment-skipping logic added in this PR (to `static/install.sh` and `static/install.ps1`), the next run of this script will:

1. Overwrite the comment-aware install scripts with upstream versions that reject `#` lines
2. Leave manifests with `##` license headers (written by the Java generator)
3. Break the install flow for users

The upstream `parse_manifest()` has `fail "manifest has unknown key: $key"` for any line that doesn't match the 4 known keys, and the upstream `install.ps1` strictly expects exactly 4 lines.

davsclaus (changes requested):
Thank you for this contribution — consolidating the manifest logic into a single Java tool is a sound direction. However, I found two critical coordination issues that would break the install flow at the next release if merged as-is.

### 1. Vendored copy diverges from upstream (not byte-identical)

The PR description and the file banner both claim the vendored `scripts/WebsiteManifestGenerator.java` is "kept byte-identical to the upstream source" at `dsl/camel-jbang/camel-launcher/src/jreleaser/java/WebsiteManifestGenerator.java`, but it has three functional changes:

- **Added** `LICENSE_HEADER` constant (~20 lines of properties-style ASF license header)
- **Modified** `renderManifest()` to prepend `LICENSE_HEADER` to manifest content
- **Modified** `parseStrictManifest()` to add `|| line.startsWith("#")` for comment-skipping

The banner says "re-copy that file over this one, then re-apply this banner" — but a blind re-copy would lose the license-header and comment-skipping changes, not just the banner.

**Suggestion**: make these changes upstream first (in apache/camel), then vendor the truly byte-identical file.

### 2. `install.sh` / `install.ps1` changes will be overwritten on next run

`update-camel-cli-manifest.sh` (step 1) still downloads `install.sh` and `install.ps1` from upstream `apache/camel main` and overwrites the local copies. The upstream versions do **not** have the comment-skipping logic added by this PR.

So the next time the script runs for a new release:
1. Local install scripts (with comment handling) are overwritten by upstream versions (without)
2. The Java generator writes manifests WITH ASF license headers (`## ...` lines)
3. The upstream `parse_manifest()` in `install.sh` hits `fail "manifest has unknown key"` on the `#` lines, and the upstream `install.ps1` fails the exactly-4-lines check
4. → **Broken user installs**

### Recommendation

The changes should land **upstream first** (in apache/camel) so that:
- `install.sh` and `install.ps1` already handle `#` comment lines when fetched
- `WebsiteManifestGenerator.java` already includes the license-header logic when vendored
- The vendored copy is then truly byte-identical, and the maintenance model works

Is there a companion PR on `apache/camel` for these upstream changes?

---

ammachado:
Companion upstream PR: apache/camel#25030. It lands all of these in apache/camel `main`, which resolves both coordination points here. Sequencing matters though: #25030 needs to merge before this PR (or before the next `update-camel-cli-manifest.sh` run), otherwise the divergence you flagged is real.

On the byte-identical concern: #25030 adds the three pieces to the upstream generator that were previously only vendored here — the `LICENSE_HEADER` constant, `renderManifest()` prepending it, and the `parseStrictManifest()` `#`-skip. Once it merges, `scripts/WebsiteManifestGenerator.java` is byte-identical to the upstream source again (bar the vendored-copy banner), so the "re-copy, re-apply banner" model stops silently dropping changes.

On the installer clobbering: #25030 also adds the `#`-comment tolerance to upstream `install.sh` (`case \#*) continue`) and `install.ps1` (`Where-Object { -not $_.StartsWith('#') }`). So after it merges, step 1 of `update-camel-cli-manifest.sh` fetches installers that already skip the `##` license header the Java generator writes. I also synced `static/install.sh` here with a small upstream robustness change (surfacing wget/tar errors) to keep the vendored copy identical to what #25030 publishes.

Given the ordering dependency, does it make sense to hold this PR until #25030 is merged?

_Claude Code on behalf of ammachado_

scripts/WebsiteManifestGenerator.java
ammachado: Addressed upstream by apache/camel#25030, which adds the `LICENSE_HEADER`, the `renderManifest()` prepend, and the `parseStrictManifest()` `#`-skip to the source file — so once it merges the vendored copy is byte-identical again (bar the banner) and a re-copy stops dropping changes. More in the top-level comment. _Claude Code on behalf of ammachado_

update-camel-cli-manifest.sh
ammachado: apache/camel#25030 adds the `#`-comment tolerance to upstream `install.sh`/`install.ps1`, so once it merges this step re-fetches comment-aware installers and the clobbering goes away. The dependency is ordering: #25030 first. _Claude Code on behalf of ammachado_

ammachado:
Latest commit (c244c6d0): `update-camel-cli-manifest.sh` step 1 now also fetches `WebsiteManifestGenerator.java` from apache/camel main and reapplies the "VENDORED COPY" banner before writing it to `scripts/WebsiteManifestGenerator.java`, so the generator stays mechanically in sync instead of needing a manual copy-paste on each update.

davsclaus (approved):

davsclaus:
@ammachado should we wait for PR https://github.com/apache/camel/pull/25030

ammachado:
@davsclaus No need to wait — all the changes from #25030 are already in this PR.

- `static/install.sh`, `static/install.ps1`, and `scripts/WebsiteManifestGenerator.java` (bar its "vendored copy" banner) are now byte-identical to the current head of #25030 (`a539ed7`). I re-synced `install.ps1`'s PowerShell 7 compatibility fixes (the `-SkipCertificateCheck` seam and the `ExternalAttributes` unsigned-cast fix) in addition to the `install.sh` wget/tar error surfacing already picked up earlier.
- The manifest-generator ASF license header pieces (`LICENSE_HEADER`, `renderManifest()`, the `#`-skip in `parseStrictManifest()`) landed separately via #25072, which this PR already synced.

So there's no remaining divergence to wait on. #25030 can still merge on its own schedule for the CI validation work — this PR doesn't depend on it landing first anymore.

davsclaus:
ok thanks

PR #1729, opened 2026-08-13
Build: gulp 5 / Antora 3.1.15 / Yarn 4.18, DocSearch v5, and UI preview tooling

ammachado:
## What this does

Five related build and UI workstreams, each independently reviewable:

1. **gulp 5 + Antora 3.1.15** upgrade, plus a silent bug found along the way.
2. **Yarn 4.1.0 to 4.18.0** with tightened supply-chain settings.
3. **Search migrated from a hand-rolled Algolia client to DocSearch v5.**
4. **UI fixes** found while checking the new search widget in a browser: admonitions, pagination, and a Camel theme for DocSearch.
5. **Preview and test tooling**, including the fix for why item 4's admonition bug looked worse than it was.

---

# 1. gulp 5 and Antora 3.1.15

### The UI bundle was being built empty

`antora-ui-camel/gulp.d/tasks/build.js` is an `async` function that returned a stream. gulp settles an async task on its promise, and a stream is not a thenable, so `bundle:pack` zipped a directory that had not been written yet.

| | `Finished 'build'` | `ui-bundle.zip` |
|---|---|---|
| before | 26 ms | 22 bytes, 0 files |
| after | 1.92 s | 697689 bytes, 124 files |

This did not affect the published site: `antora-playbook-production.yml` consumes the staged `antora-ui-camel/public/_` directory, not the zip. It did affect anything using the `--ui-bundle-url` that the gulp task prints.

### Theme to gulp 5

Moved to gulp 5, vinyl-fs 4, undertaker 2, through2 5 and fs-extra 11. Three things needed more than a version bump:

- **`merge-stream` had to go.** vinyl-fs 4 is built on `streamx`, and `merge-stream` is a node-core `PassThrough` that silently drops streamx sources. Reproduced standalone: 93 files in, 3 out. Replaced with `ordered-read-streams`, the streamx-native merge from the same org as gulp.
- **Binary assets need `encoding: false`,** since gulp 5 decodes contents as UTF-8 by default. All 58 fonts and images were checked byte-for-byte against the gulp 4 output and are identical.
- **Removed the `font/*.{ttf,woff*(2)}` glob.** `src/font` has never existed in this repo's history; fonts are copied into `dest/font` by the `postcssUrl` handler. vinyl-fs 3 ignored the missing directory, vinyl-fs 4 raises ENOENT.

### PostCSS 7 plugin API to PostCSS 8

`cssnano@8` returns a plugin object rather than a callable, and the bare `(css, result) => ...` plugins were PostCSS 7 style and were no longer being invoked at all. Rewritten as visitor plugins. 97 transitive packages pruned.

`postcss-custom-properties@15` keeps the `:root` block after substitution where v9 removed it, so resolved custom properties are now dropped explicitly by a small local plugin (`gulp.d/lib/drop-resolved-custom-properties.js`).

Note for reviewers: `browserslist: last 2 versions` resolves to include `ie 11`, `ie 10`, `ie_mob` and `op_mini all`, none of which support CSS custom properties. Dropping `postcss-custom-properties` entirely is therefore not an option; it would be both larger and broken on those targets.

### Root gulpfile to gulp 5

Replaced `del` with native `fs.rm` (del 7+ is ESM only) and dropped the dependency. Fixed `generate-markdown` signalling completion twice, by both calling `done()` and returning a promise.

### Antora 3.1.15

Antora 3.1.9 added a heuristic that warns when an extension's `register` function names its first parameter `registry`:

```js
const ASCIIDOCTOR_REGISTER_FUNCTION_RX = /^(?:(?:function(?: +register)? *)?\( *registry *[,)])/
```

`@djencks/asciidoctor-jsonpath` and `@djencks/asciidoctor-antora-indexer` both do. Antora still registers them, the warning is advisory, but this playbook sets `runtime.log.failure_level: warn`, so the warning alone failed the build. Both packages were last published in 2022 and cannot be fixed upstream, so they are now required through thin wrappers in `extensions/`, matching the three local extensions already there.

Worth knowing if anyone pins differently: **3.1.11 skips these extensions entirely** rather than warning. 3.1.12 through 3.1.15 restored warn-and-register.

### Unrelated CSS fix

Removed `max-width: var(--static-max-width)` from `.blog`. That variable was never declared, only `--static-max-width--desktop` is, so the declaration was invalid at computed-value time and `max-width` already resolved to `none`. Removing it is a no-op and matches `.static`, which constrains width only inside the desktop media query.

---

# 2. Yarn 4.1.0 to 4.18.0

Fixes the stale `compat/typescript` patch-hunk warning during install: the patch hash no longer matched typescript 5.9.3 under the old Yarn.

Supply-chain settings tightened at the same time:

- `approvedGitRepositories: []`, since this repo has no git-protocol dependencies.
- `npmMinimalAgeGate: 4320` (3 days), delaying trust in newly published npm versions.

# 3. Search: hand-rolled Algolia client to DocSearch v5

The site previously shipped `src/js/vendor/algoliasearch.bundle.js`, a 365-line hand-rolled search UI built on the raw `algoliasearch` v4 client, along with 169 lines of bespoke result-dropdown CSS in `src/css/header.css`. Both are removed in favor of `@docsearch/js` 5 and `@docsearch/css` 5, which is the officially supported integration for the DocSearch-crawled index this site already uses.

### Result curation had to be carried across

The old bundle did three things beyond querying, and a straight swap to DocSearch would have silently dropped all of them. They are reimplemented in `src/js/08-docsearch.js`:

- **Sub-project exclusion.** `/camel-k/`, `/camel-quarkus/`, `/camel-spring-boot/`, `/camel-kafka-connector/`, `/camel-kamelets/` and `/camel-karaf/` stay out of the main index results; those docs are browsable directly.
- **Core-docs ranking**, floating `/manual/`, `/user-guide/`, `/architecture/`, `/getting-started/` and `/faq/` above `/components/`. This works through `transformItems` because DocSearch groups by `hierarchy.lvl0` into an insertion-ordered plain object, so the order we emit decides both group order and within-group order.
- **Per-page capping** at 2 hits, replacing the old query-aware `deduplicateHits`. `transformItems` never receives the query, and the original only used it to detect a direct parent match, so capping by page reaches an equivalent outcome without it.

### Why `hitsPerPage: 75`

The index has no `attributeForDistinct` and no `attributesForFaceting`, so both filters must run client side, on an already-truncated window of hits. Distinct pages surviving the sub-project filter, measured against the live index, at 75 vs. the DocSearch default of 20: `timer` goes from 5 to 8 pages, `aggregate` from 9 to 13. 100 adds almost nothing beyond that but doubles the response to ~36 KB gzipped.

At the default, `timer` returned few usable hits spanning a single document, so the first result group was several near-identical anchors off one page.

The durable fix is at the index level rather than in the browser. Filed as **[CAMEL-24396](https://issues.apache.org/jira/browse/CAMEL-24396)**; with server-side `distinct` the over-fetch and most of the client-side logic can be deleted.

Note that `.docsearch.config.json` in this repo already attempts both settings, but they are not in effect on the live index, and one of them (`attributeForDistinctResults`) is not a real Algolia setting name. The records also lack `url_without_anchor`, so there is currently no anchor-free attribute to group on. Details in the ticket; nothing there blocks this PR.

### `@docsearch/css` gets its own build stream

Routing the vendor stylesheet through `postcss-custom-properties` with `preserve: false` flattened its runtime cascade: the `@media (width<=768px)` `:root` overrides were inlined at desktop values, `--shimmer-bg` was dead, and the 2.5 KB `:root[data-theme=dark]` block still shipped but could never take effect.

It now builds separately, skipping the custom-property pass, revved to `css/vendor/docsearch.css` and linked ahead of `site.css` in both the Antora and Hugo head partials. This mirrors how the DocSearch UMD bundle already ships as vendor JS.

`preserve: true` was rejected deliberately: it fixes the same bug but costs about 3.8 KB gzip site-wide and would reverse a flattening convention that predates this branch. Total CSS actually shrank slightly, since DocSearch's properties are no longer duplicated by inlining.

---

# 4. UI fixes found while checking the theme in a browser

### Admonitions lost their icon column

`src/css/doc.css` set `table-layout: fixed` on `.doc .admonitionblock > table`, inherited from the default Antora UI. That UI positions `td.icon` absolutely, so collapsing the icon column costs it nothing. This UI keeps the icon in flow as a vertical label bar, and under fixed layout it gets zero width because `td.content` is `width: 100%`.

Removing the one declaration is a production no-op, confirmed by diffing the compiled stylesheet rule by rule: exactly one declaration differs. The rule matches nothing on the built site anyway, since `extensions/table.js` wraps every table in a `div.table-wrapper` and breaks the child combinator. It only ever bit the preview, which did not run that postprocessor. See item 5.

### Pagination links drew a full-width underline

`nav.pagination span` stretched its anchor to half the container. `.doc a` draws its underline with a repeating background-image gradient, so a stretched anchor painted that dashed line across the whole box rather than under the label. Fixed with `align-items` so the anchor keeps its text width.

Bundled alongside that fix is a small visual redesign of the same component: `gap` instead of margin-based spacing, a `border`/`border-radius` around each link, and a hover accent, replacing the old top separator.

### DocSearch widget theme

The widget shipped in Algolia's default blue. It is now themed in `src/css/docsearch.css`, scoped to `.DocSearch`, which is present on both the launch button and the modal container.

Two traps are worth flagging for reviewers, both documented in the file:

- **Values are literal, not `var()` references.** `postcss-custom-properties` substitutes `var()` in ordinary declarations but not inside custom-property declarations, and the `:root` palette is stripped from the output. A first version written as `var(--color-camel-orange)` compiled to an unresolved `var()` against a deleted variable, so the theme was silently dead.
- **Seven derived vendor properties are restated.** DocSearch derives them at `:root`, so they resolve at their declaration site and overriding the inputs lower in the tree does not propagate.

Three vendor properties are deliberately not set, because the corresponding SVGs hardcode their colors in inline `<style>` blocks. `--docsearch-logo-color` was dropped rather than ship dead code, and the no-results glyph is handled with a `stroke: currentcolor` rule instead.

Also sized the searchbox in `rem`: the vendor default is a fixed `56px`, which clipped descenders in the input against this site's 18px root font size. Ruled out the font swap as the cause by toggling to `system-ui` in the live page, where the text still measured 19px.

Four unit tests cover the theme, all mutation-checked: scoped to `.DocSearch` rather than `:root`, no `var()` left in custom-property declarations, searchbox height in `rem`, and the derived properties restated.

---

# 5. Preview and test tooling

### The preview diverged from the built site

`gulp preview` ran plain Asciidoctor with none of the postprocessors `antora-playbook-production.yml` registers, so what it rendered was not what the site rendered. That is what made the admonition bug in item 4 look like a theme regression when it was a harness artifact. Measured across four environments: local preview reported `table-layout: fixed` with a 0px icon column, while camel.apache.org, the Netlify deploy preview and the locally built site all reported `auto` and 33px.

`build-preview-pages.js` now registers the same extensions as the playbook. `@asciidoctor/tabs` is deliberately left out and the reason is documented inline: the 1.0.0-beta.3 build targets the Opal runtime of asciidoctor.js 2 and throws on require under the `@asciidoctor/core` 3 this UI depends on. The site build is unaffected, since Antora supplies its own Asciidoctor, so tabs render as description lists in the preview only.

### Example pages

`preview-src` gained pages for admonitions, source code, tables, tabs and asciinema, wired into the sample nav. `tables.adoc`, `asciinema.adoc` and their content are original; `404.adoc`, `admonitions.adoc`, `code.adoc`, `index.adoc`, `tabs.adoc`, `ui-model.yml` and `multirepo-ssg.svg` are adapted from [bonitasoft/bonita-documentation-theme](https://github.com/bonitasoft/bonita-documentation-theme), MPL-2.0 and therefore ASF Category B — the same license `antora-ui-camel` already ships under. Attribution is in `README.adoc` and inline in the most directly copied files.

### Tests run from gulp

`gulp test` uses `run()` from `node:test` so gulp settles on the returned promise and a failing test fails the task it is sequenced into, which is `bundle`. Verified both ways: exit 1 with a forced failure, exit 0 when restored.

### asciinema

`@springio/antora-extensions/asciinema-extension` is registered in the playbook, with the player assets and partials guarded on the page attribute it sets, so pages without a cast load neither the script nor the stylesheet.

The block is an Antora extension and never runs under `gulp preview`, so the preview reproduces it locally against the extension's own partials and helpers from `node_modules`. One subtlety worth knowing if this is ever touched: the cast ids are collected into a plain object rather than set on the document, mirroring what the extension does with `file.asciidoc.attributes`. Asciidoctor rolls back attributes assigned from the body once parsing ends, so a `page-` attribute set from a block processor is gone by the time `getAttributes()` runs.

### Unused media detection

`extensions/detect-unused-media.js` reports unreferenced images to `build/unused-media.txt`. It logs at info rather than warn on purpose: this playbook sets `runtime.log.failure_level: warn`, so warning once per unused file would fail the whole site build over a housekeeping issue. That is why the off-the-shelf `@bonitasoft/antora-detect-unused-media-extension` was dropped in favor of a local one.

---

## Verification

- Theme build: `ui-bundle.zip` at 697689 bytes / 124 files, and the CSS hash is reproducible across runs
- All 58 binary assets byte-identical to the gulp 4 baseline
- `yarn build:antora`: exit 0, **4574 HTML files** across all nine components, indexer output intact (camel-kafka-connector 218 pages, camel-kamelets 505)
- The asciinema extension is confirmed to run in a real build: `documentation/_/js/vendor/asciinema-player.js` and `css/vendor/asciinema-player.css` exist in the output, and they get there only through its `uiLoaded` hook
- `extensions/detect-unused-media.js` wrote a 30-entry report during that build and, as designed, logged no warnings, so it did not trip `failure_level: warn`
- `sitemap`, `htaccess` and `generate-markdown` (4579 files) all pass under gulp 5
- **22 unit tests**, now run by `gulp test` and sequenced into `bundle`. They cover the custom-property plugin, the search transform and the DocSearch theme. The search suite evaluates the shipped IIFE against stubbed globals and replays captured Algolia responses, so it exercises the real file rather than a copy. Mutation-checked throughout: weakening the per-page cap, dropping to `hitsPerPage: 20`, removing the sort, or moving the theme to `:root` each fail the relevant assertions.

Checked in a browser: the themed DocSearch modal, keyboard selection and the no-results screen; admonitions across four environments (see item 5); and both asciinema players mounting and playing through in the preview.

### Still to verify

- **The `[asciinema]` block has never been exercised in a real site build.** Everything here runs against the preview reimplementation. No content source carries a cast yet, so the extension's block processor and the `_asciinema/<id>.cast` output remain unproven end to end.
- The DocSearch modal on mobile, and the `data-theme="dark"` toggle.
- The demo cast is a `camel tui` session recorded at 160x44. At the width of the docs content column that is small text, so the size may want revisiting.


ammachado:
## Code review follow-up (d84cef15)

A review pass over the DocSearch v5 migration turned up six findings; all are addressed in `d84cef15`. Summarizing the two that changed behavior, since they are worth a look rather than just a rubber stamp.

### Search result curation was silently dropped

The deleted `src/js/vendor/algoliasearch.bundle.js` filtered sub-project hits, de-duplicated hits pointing at the same parent page, and ranked core docs above component pages. The v5 config carried none of that over. Restored in adapted form:

- **Sub-project exclusion** via `transformItems`.
- **Core-docs ranking** via `transformItems`. This works because DocSearch groups by `hierarchy.lvl0` into an insertion-ordered plain object (`buildQuerySources` in `@docsearch/react`), so the order we emit decides both group order and within-group order.
- **Per-page cap** of 2 hits, replacing the old query-aware `deduplicateHits`. `transformItems` never receives the query, and the original only used it to detect a direct parent match, so capping by page reaches an equivalent outcome without it.
- **`hitsPerPage: 50`** instead of the DocSearch default of 20.

That last one is a real regression fix, not tuning. Both filters run client side, so they operate on an already-truncated window. Distinct parent pages surviving the sub-project filter, measured against the live index:

| query | hitsPerPage 20 | hitsPerPage 50 |
|---|---|---|
| kamelet | 2 pages | 3 pages |
| timer | **1 page** | 7 pages |
| rest dsl | 7 pages | 16 pages |

At the default, `timer` returned 19 usable hits spanning a single document, so the first result group was five near-identical anchors off one page.

### Why this is worked around in the browser at all

The durable fix belongs at the index level, and there is more going on there than is obvious. Filed as **[CAMEL-24396](https://issues.apache.org/jira/browse/CAMEL-24396)**. Summary of what I verified against the live index:

- **`.docsearch.config.json` already tries to configure both settings** (`"attributesForFaceting": ["version"]` and `"attributeForDistinctResults": "url"`), but **neither is in effect**. Requesting `{"facets": ["version"], "hitsPerPage": 0}` returns `"facets": {}`, and `{"distinct": true}` is a no-op.
- **`attributeForDistinctResults` is not a real Algolia setting.** The name is `attributeForDistinct`, and it must be paired with `distinct`, which is absent. Per Algolia's API reference: "If `attributeForDistinct` isn't set, `distinct` is ignored."
- **The file matches neither documented DocSearch config format.** The current DocSearch/Crawler config is JavaScript (`new Crawler({ startUrls, sitemaps, actions, initialIndexSettings })`), not JSON nested under `index` / `crawler` / `custom_settings`. Nothing in this repo references the file outside `README.md` and `.docsearch.README.md`; there is no CI job or script that applies it.
- **The records lack the attribute the fix would normally use.** Retrieving all attributes returns only `content, hierarchy, keywords, objectID, pageRank, url, version`. There is no `url_without_anchor`, `anchor`, `type` or `weight`, so this index is not built by DocSearch's standard `helpers.docsearch()` extractor. Since `url` includes the `#fragment`, `attributeForDistinct: "url"` would group by anchor and de-duplicate nothing.

So the client-side workaround is currently the only lever available, and the ticket's first step is confirming whether `.docsearch.config.json` feeds the crawler at all or whether the live config lives only in the Algolia dashboard.

One thing worth watching during review: DocSearch v5 uses `item.type === 'lvl1'` to nest sub-results under their parent heading. With no `type` attribute on these records, that grouping cannot engage. Search works, since `hierarchy`, `url` and `content` are all present, but the result list will not render DocSearch's usual parent/child structure.

### `@docsearch/css` lost its runtime cascade

The vendor stylesheet was going through `postcss-custom-properties` with `preserve: false`, which flattened it. Verified in the built output: the `@media (width<=768px)` `:root` overrides were inlined at desktop values, `--shimmer-bg` was dead, and the 2.5 KB `:root[data-theme=dark]` block still shipped but could never take effect.

`@docsearch/css` now gets its own build stream that skips the custom-property pass, revved to `css/vendor/docsearch.css` and linked ahead of `site.css` in both the Antora and Hugo head partials. This mirrors how the DocSearch UMD bundle already ships as vendor JS.

I rejected the one-line `preserve: true` alternative deliberately: it fixes the same bug but costs about 3.8 KB gzip site-wide and would reverse a flattening convention that predates this branch. Total CSS actually shrank slightly, since DocSearch's properties are no longer duplicated by inlining.

### Also fixed

- The new `drop-resolved-custom-properties` plugin was deleting `@media`-nested `:root` definitions that `postcss-custom-properties` never substitutes (exactly the block `@docsearch/css` ships), while missing resolved `:root, :host` list forms.
- The `@docsearch/js` UMD path now fails loudly instead of surfacing as an opaque vinyl-fs glob error.
- Added a `window.docsearch` guard so a vendor-bundle load failure cannot take down the rest of the `site.js` listener chain.

### Tests

16 cases, all passing. The search-transform suite evaluates the shipped IIFE against stubbed globals and replays captured Algolia responses, so it covers the real file rather than a copy. Mutation-checked: weakening the per-page cap, dropping to `hitsPerPage: 20`, or removing the sort each fails the relevant assertions.

### Still to verify

Nothing has been exercised in a browser yet. Keeping this as a draft until the DocSearch modal on mobile, the `data-theme="dark"` toggle, and the `timer` / `kamelet` result lists are checked against a running site.

---

*Edited to correct the index-level section: an earlier revision stated the fix could not be done from this repository, which was wrong. `.docsearch.config.json` exists; the accurate finding is that its settings are not in effect and one of them is misspelled.*

antora-ui-camel/src/js/08-docsearch.js
```diff
@@ -0,0 +1,128 @@
+;(function () {
+  'use strict'
+
+  // The index has no attributeForDistinct and no attributesForFaceting, so both the sub-project
+  // exclusion and the per-page dedupe below have to run client side. Fetch deeper than the
+  // DocSearch default of 20 so filtering does not starve the result list: at 20 a query like
+  // "timer" is left with hits from a single page. Measured against the live index, 75 lifts
+  // "timer" from 5 to 8 distinct pages and "aggregate" from 9 to 13; 100 adds almost nothing
+  // beyond that but doubles the response to ~36 KB gzipped.
+  var HITS_PER_PAGE = 75
```
davsclaus, line 10: The PR description's table says `hitsPerPage: 50`, but this sets `HITS_PER_PAGE = 75`. Not a bug — just stale prose in the description worth a quick correction. Also worth noting: the new test only asserts `>= 50`, so it wouldn't catch a future regression back down to 50.

antora-ui-camel/src/css/pagination.css
```diff
@@ -1,31 +1,40 @@
...
 
+/* NOTE align-items keeps the anchor at its text width. .doc a draws its underline with a
+   repeating background-image gradient, so a stretched anchor paints that dashed line across the
+   whole half-width of the box instead of just under the label. */
 nav.pagination span {
   display: flex;
-  flex: 50%;
+  flex: 0 1 calc(50% - 0.5rem);
   flex-direction: column;
+  gap: 0.4rem;
```
davsclaus, line 15: The PR description frames this file's change as a one-line `align-items` fix for the underline-stretching bug, but this hunk also adds `gap`, `border`, `border-radius`, and a hover accent, and removes the old top separator. The underline fix itself (`align-items` on `.prev`/`.next`) is correct — just flagging that a small visual redesign is bundled in beyond what the description calls out.

tools/asciinema/.camel-jbang/camel-jbang-run.properties
```diff
@@ -0,0 +1,26 @@
+loggingLevel=info
```
davsclaus: This looks like a JBang-generated local run-config dotfile (auto-created under `.camel-jbang/` when running `camel-jbang`). No secrets and nothing machine-specific, but worth confirming this is meant to be a committed fixture for reproducing the demo recording rather than an accidental local artifact.

davsclaus (approved):
Reviewed against this project's build/CI conventions and general code-quality expectations (not a substitute for CodeRabbit/Sourcery/SonarCloud or a dedicated security audit).

## Summary

Large (76-file) but unusually well-documented build/infra PR — gulp 5 + Antora 3.1.15, Yarn 4.18, a DocSearch v5 migration, several UI fixes, and new preview/test tooling. Nearly every non-obvious change has an inline comment explaining the "why," and the PR body's claims check out against the diff. **No blocking issues found.**

### Worth highlighting (positive)
- The old hand-rolled `algoliasearch.bundle.js` rendered Algolia hit data via `innerHTML` with the raw query — a reflected-XSS-shaped pattern. Moving to `@docsearch/js`'s Preact rendering removes this as a side effect of the migration.
- The gulp 5 async/stream bug (empty `ui-bundle.zip`) was real and is correctly fixed.
- The `indices: [{name, searchParameters}]` DocSearch config shape looked suspicious at first (resembles a legacy v1/v2 API) but checks out against the actual `@docsearch/react@5.0.0` type definitions — it's the documented v5 shape.
- Clean dead-code removal (old search CSS/JS, a stray years-old `.build.js.swp` file).

### Non-blocking notes (see inline comments for specifics)
1. `antora-ui-camel/src/img/cancel.svg` looks orphaned: the cancel-button `` referencing it was removed from `header-content.hbs` (both `src/` and `public/_/` copies), but the SVG file and its `asset.js`/rev-manifest entries were left behind. Ironically, the PR's own new `detect-unused-media.js` extension should catch this on the next build.
2. The `[asciinema]` Antora extension block has never been exercised in a real site build (per the PR's own "Still to verify" section) — no content uses it yet. It's additive/opt-in so it can't regress existing pages, but worth verifying end-to-end (or landing one real page using it) before or shortly after merge.
3. `test/record-cast-test.js` spawns `python3` — worth confirming CI images have it available.
4. Whether the new `preview-src` example pages are genuinely original content vs. adapted from an MPL-2.0 theme is a licensing call (ASF Category B) worth a maintainer's own judgment.

antora-ui-camel/package.json
```diff
@@ -9,41 +9,42 @@
     "url": "https://github.com/apache/camel-website.git"
   },
   "engines": {
-    "node": ">= 8.0.0"
+    "node": ">= 18.0.0"
```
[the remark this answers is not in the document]
ammachado: Fixed in c4c0b885: dropped `through2` (all five call sites only ever wanted an object-mode Transform, which `node:stream` provides directly via a new `gulp.d/lib/map.js` helper) and raised `engines.node` to `>= 22.12.0` to match `.nvmrc`. Node 18/20 are both EOL, so no need to keep supporting them.

antora-ui-camel/src/js/08-docsearch.js
ammachado: Good catch, tightened the test to `assert.equal(searchParameters.hitsPerPage, 75, ...)` so a regression to 50 (or any other value) now fails it. Leaving the PR description as-is for the maintainer to reconcile, since that's prose rather than code.

antora-ui-camel/src/css/pagination.css
ammachado: Correct, that's intentional — I bundled a small visual redesign (gap, border, radius, hover accent) along with the underline fix rather than splitting it out. Will tighten the PR description to call that out explicitly rather than changing the diff.

ammachado:
Follow-up on the non-inline notes from the review:

**Asciinema extension unexercised** — confirmed, no content uses `[asciinema]` yet. It's additive/opt-in so it can't regress anything today; landing a real page that uses it (or at least an end-to-end manual check) is tracked as a follow-up rather than blocking this PR.

**Test suite not exercised in CI** — digging into this, the gap is actually broader than the `python3`-in-CI question: neither the root `yarn test` (which includes `record-cast-test.js`) nor `antora-ui-camel`'s `gulp test` (the DocSearch suite, etc.) is wired into `.github/workflows/pr.yaml` at all — the `checks` job only runs `check:links`, `check:html`, `check:redirects`. So `python3` availability is moot for now. Wiring the test suites into CI is worth doing but is scoped as a separate follow-up rather than expanding this PR further.

**Licensing of `preview-src`** — confirmed: several of the new example pages (`404.adoc`, `admonitions.adoc`, `code.adoc`, `index.adoc`, `tabs.adoc`, `ui-model.yml`, `multirepo-ssg.svg`) are adapted from [bonitasoft/bonita-documentation-theme](https://github.com/bonitasoft/bonita-documentation-theme), which is MPL-2.0 — the same license `antora-ui-camel` already ships under (inherited from the Antora default UI since this subproject's creation). Added attribution in `d84cef15`-adjacent commit: a note in `antora-ui-camel/README.adoc`'s Copyright and License section, plus inline comments in `404.adoc`, `ui-model.yml`, and `multirepo-ssg.svg` (the most directly copied files) pointing back to the source.

tools/asciinema/.camel-jbang/camel-jbang-run.properties
ammachado: It was an accidental local artifact, not meant to be a committed fixture. Removed.

davsclaus:
LGTM

PR #1749, opened 2026-09-12
chore: remove redesign planning docs, drop dangling CSS comment refs, regen UI bundle

ammachado:
## Summary

- **Remove redesign planning docs.** Deletes the specs and implementation plans under `docs/superpowers/` (antora docs UI, foundation tokens and typography, shared chrome, home page). They were working notes for the new website design, which landed in #1740.
- **Drop dangling CSS comment pointers.** Eight comments in `antora-ui-camel/src/css/` (`blog.css`, `category.css`, `primitives.css`, `release.css`, `security.css`, `tooling.css`) cited `.superpowers/sdd/4b/` design notes (`00-primitives.md`, `extract/*.md`) that were never committed. The pointers are removed; each comment keeps the reasoning it already stated inline. Comment-only.
- **Regenerate `yarn.lock` and the UI bundle.** `yarn.lock` was regenerated from scratch, which moves dependencies to their latest in-range versions, and `antora-ui-camel/public/_` was rebuilt with `yarn workspace antora-ui-camel run build`. The bundle gets new hashed site/vendor CSS and JS, updated fonts, and matching `rev-manifest` and partial references.
- **Keep CI's deadlinks checksums.** The three `@deadlinks/cargo-deadlinks*` archives are built locally by the `github-release-binary` plugin, and the ones built on macOS did not match what CI builds (`YN0018`). Their checksums are restored to the #1747 values, which CI accepted on a fresh fetch. The UI build does not use deadlinks.

## Notes for reviewers

- The bundle diff comes entirely from the lockfile refresh. Building with and without the CSS comment edits on the same dependencies produced byte-identical output (114 files).
- The bundle was built locally with Node 26, while CI and the regen workflow use Node 22 (`.nvmrc`). If the output differs between the two, the regen workflow will open a follow-up PR after merge.
- Fresh deadlinks archives built on macOS still do not match the committed checksums, so a full `yarn install` on macOS can fail with `YN0018`. That predates this PR and is worth tracking separately.

## Test plan

- [x] Nothing in the repo references `docs/superpowers/` (`git grep -n superpowers`).
- [x] No `.superpowers/` or `00-primitives` citations remain under `antora-ui-camel/src`.
- [x] The CSS comment edits do not change the built bundle (same-dependency builds, byte-identical).
- [x] The deadlinks lockfile entries match 5a994408 exactly.
- [ ] PR checks pass (`yarn workspaces foreach --all install`, `yarn build-all`, `yarn checks`).

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Adriano Machado (@ammachado)

_This was generated by an AI agent and may contain inaccuracies. Please verify before relying on it._


davsclaus (approved):

oscerd (approved):

davsclaus:
@ammachado the carusel with the 3 kafka examples show the same file name in the bottom

$camel run kafka-to-db.yaml

we need to adapt this to the different DSL

PR #1780, opened 2026-09-19
Add blog post Extending Apache Camel Simple with Your Own Functions

SiHartl:
This PR adds a blog post about creating simple custom functions in Apache Camel.

The article demonstrates how custom functions can be implemented and used in Camel expressions.


oscerd (approved):

davsclaus (changes requested):
Thanks for the post, nicely written. I verified the technical content against the Camel source (`SimpleFunction` API, `$init{}` / `~:=` syntax, `@BindToRegistry` discovery, all reference links) and ran the two-file example locally.

One thing I'd ask to change before we merge:

**Camel 4.22.1 was released on 2026-09-17 and contains the fix for CAMEL-24486.** I ran the exact example with `camel run --camel-version=4.22.1 route.camel.yaml MaskEmailFunction.java` in the default `dev` profile and it works without any workaround:

```
Processed customer:
Customer: JOHN DOE
Contact: j***@example.com
```

The post is currently written against 4.22.0 and mentions the `--profile test` workaround in about six places (intro, quick start step 4, Step 4, Expected result, Complete runnable example), and the very first command a reader sees is the workaround one. Since the post will be published after 4.22.1 is out, could you:

- state that the example needs Camel 4.22.1 or newer,
- make `camel run route.camel.yaml MaskEmailFunction.java` the primary command everywhere, and
- collapse the workaround to a single short note along the lines of "if you are on 4.22.0, add `--profile test` because of CAMEL-24486".

Non-blocking nits:

- The expected output shows `Name:   John   Doe`, but the log actually prints `Name: John   Doe` (leading whitespace of the header is not visible in the log line).
- `index.md` is missing a trailing newline.

SiHartl:
Thanks for the review.

I've updated the post to require Camel 4.22.1+, changed the examples to use the normal camel run command, and reduced the Camel 4.22.0 workaround to a single note.

I also addressed the additional nits.

Could you please have another look?

davsclaus (approved):

davsclaus:
Thanks

